PR #5, opened 2015-02-04
Make consul auto configuration conditional

zachmarshall:
In some cases it's desirable to disable Consul auto configuration. For example, when running a service that uses spring-cloud-consul-config, you currently have to have a Consul server running or the application context fails to load. It would be nice to be able to disable the Consul auto configuration in these cases.

This PR uses `@ConditionalOnProperty(name = "consul.enabled", matchIfMissing = true)` to trigger creation of the various Consul components.

### DIFF
--- a/spring-cloud-consul-core/src/main/java/org/springframework/cloud/consul/ConsulAutoConfiguration.java
+++ b/spring-cloud-consul-core/src/main/java/org/springframework/cloud/consul/ConsulAutoConfiguration.java
@@ -2,6 +2,7 @@ package org.springframework.cloud.consul;
 
 import com.ecwid.consul.v1.ConsulClient;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
@@ -11,6 +12,7 @@ import org.springframework.context.annotation.Configuration;
  */
 @Configuration
 @EnableConfigurationProperties
+@ConditionalOnProperty(name = "spring.consul.enabled", matchIfMissing = true)
 public class ConsulAutoConfiguration {
 
     @Bean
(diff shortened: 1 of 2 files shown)


zachmarshall:
Also, I work for SAS Institute and we've signed the contributor agreement, so if this PR is satisfactory you should be safe to pull it. If it's missing anything or you prefer a different approach, please let me know and I may be able to implement it.

spring-cloud-consul-bus/src/test/java/org/springframework/cloud/consul/bus/ConsulBusIT.java
```diff
@@ -6,6 +6,7 @@
 import org.junit.FixMethodOrder;
 import org.junit.Test;
 import org.junit.runners.MethodSorters;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
```
spencergibb: Not needed in this file as this is just a test

spring-cloud-consul-core/src/main/java/org/springframework/cloud/consul/ConsulAutoConfiguration.java
```diff
@@ -11,6 +13,7 @@
  */
 @Configuration
 @EnableConfigurationProperties
+@ConditionalOnProperty(name = "consul.enabled", matchIfMissing = true)
```
spencergibb: I think we should do `spring.cloud.consul`.  We would need to make sure it is consistent throughout the project.

zachmarshall: Thanks for the feedback. This was intended to piggyback off the ConsulProperties class. I guess the plan there is to update to `spring.consul` as the prefix?

spencergibb: yeah, we started moving that way, hadn't made it here yet.

zachmarshall:
Would you prefer I submit this on master? I picked this branch because at the time this was where you were working, but I see you're moving master forward, too.

spencergibb:
yeah, master would be better.  `consul-client` is me replacing my own partial consul client with a more fully featured one.

zachmarshall:
Closing in favor of #7 against branch master

PR #7, opened 2015-02-05
Make consul auto configuration conditional

zachmarshall:
In some cases it's desirable to disable Consul auto configuration. For example, when running a service that uses spring-cloud-consul-config, you currently have to have a Consul server running or the application context fails to load. It would be nice to be able to disable the Consul auto configuration in these cases.

This PR uses `@ConditionalOnProperty(name = "spring.cloud.consul.enabled", matchIfMissing = true)` to trigger creation of the various Consul components.

Replaces PR #5

### DIFF
--- a/spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulConfigBootstrapConfiguration.java
+++ b/spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulConfigBootstrapConfiguration.java
@@ -17,6 +17,7 @@
 package org.springframework.cloud.consul.config;
 
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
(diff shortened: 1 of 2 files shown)


zachmarshall:
I've been thinking about this some more. This PR was created to handle the case where consul isn't running and you don't want startup to fail (eg integration tests). My recent thoughts have me wondering: should spring-cloud-consul silently proceed (and not fail startup) if it can't find a consul instance at the consul url instead of using a property to turn it off? What do the other spring-cloud providers do when the thing they want to communicate with is not there?

aatarasoff:
Hi, Zach. I think, you should use profiles to rule needed behavior.

aatarasoff:
Imagine that consul is unreachable in production environment. Do you think ignoring this fact is better than fail on startup?

spencergibb:
@zachmarshall I generally like things to be disabled.  Eureka (from netflix), does registration in a background thread.  If consul fails to connect, there should be an option for the service to either continue to run, or retry a certain number of times (we have that for config server).

PR #12, opened 2015-03-11
List service nodes filtered by alive, when getting the server list

nmarasoiu:
Hi,

This is a first pull request.
Most of the code is in a new module, that anyone can pull in maven if they want it.
No modification to existing modules is done, only a few additions of fields and remote http call methods.
I can also create an issue/feature request to track this.

The main functional idea is that a new server list filter is added in a separate module, but which needs few additions in existing core ones.

The list of server serving a service will be filtered by serf status, so that only services hosted by "alive" members of the gossip pools are listed.

The next would be the ttl implementation.

Thank you and cheers!
Nicu

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,6 +32,7 @@
         <module>spring-cloud-consul-bus</module>
         <module>spring-cloud-consul-sample</module>
         <module>spring-cloud-consul-tests</module>
+        <module>spring-cloud-consul-utils</module>
 		<module>docs</module>
     </modules>
 
--- a/spring-cloud-consul-core/src/main/java/org/springframework/cloud/consul/client/AgentClient.java
+++ b/spring-cloud-consul-core/src/main/java/org/springframework/cloud/consul/client/AgentClient.java
@@ -2,8 +2,10 @@ package org.springframework.cloud.consul.client;
 
 import feign.Param;
 import feign.RequestLine;
+import org.springframework.cloud.consul.model.Member;
 import org.springframework.cloud.consul.model.Service;
 
+import java.util.List;
 import java.util.Map;
 
 /**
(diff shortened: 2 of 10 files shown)


nmarasoiu:
Hi again,

I want to add something: what do you think about using an existing Consul Java client, such as Orbitz or Ecwid? They are more up to date than cloud consul, Orbitz works quite alot with automatic mapping as well, so not error prone, and they look like an appropriate intermediate layer. It is true indeed that adding quickly some new features from latest consul (now 0.5) may need intervention in the now-outsourced layer, but it is much more common I think (and the TTL task is more abundant on that), the case where we need to add fields and remote http methods, or correct fields which are there correct already (for instance, Check was not really mappable as such, I think interval and TTL need to be string, some id and name need introducing, and such, which are already fine in their domain model). Indeed, mapping between 2 domain models will also seem awkward.

My recommendation would be to rewrite based on an existing client, if one seems solid (I think they do), and thus use their domain model (which is similar, it is consul domain, but in a different package), and, of course, the http mappings, which are not feign, but also use automatic binding with json.

And we can keep in cloud-consul only higher level domain objects if any, composing the lower level, or containing aggregates, abstractions, or other view/usecase-oriented data.

What do you think?

spencergibb:
https://github.com/spring-cloud/spring-cloud-consul/tree/consul-client

--
Spencer Gibb
spencer@gibb.us

nmarasoiu:
Ok, I am rebasing now on consul-client. Will do a single pull for both features.
Other than client-related, I would be interested in any other feedback, at any time.
I am also working on ttl to have a server list filter on that too (on check statuses).
I will be only 1 hour per day for a few weeks, but I hope the features are in a good shape

nmarasoiu:
Also, I see you use a "0.1" ecwid version.
But ecwid does have a repo and a 1.x version: http://nexus.ecwid.com/content/groups/public/com/ecwid/consul/consul-api/1.0.8/

nmarasoiu:
However does the branch builds with tests?

spencergibb:
I didn't see their nexus repo.  I also haven't completed the migration (the event bus module).  I've been at a conference this week so I haven't had time to review your stuff.  I'll get back into it more next week.

PR #32, opened 2015-05-27
use java8 language features like lambdas

spencergibb:
Compiles to java 7.

Animal Sniffer in [spring-cloud-build](https://github.com/spring-cloud/spring-cloud-build/blob/master/spring-cloud-build/pom.xml#L251-L271) keeps java 8 api's out of use.

What do you think @dsyer @mstine @joshlong @philwebb?

### DIFF
--- a/spring-cloud-consul-sample/src/main/java/org/springframework/cloud/consul/sample/SampleApplication.java
+++ b/spring-cloud-consul-sample/src/main/java/org/springframework/cloud/consul/sample/SampleApplication.java
@@ -41,6 +41,8 @@ import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 
+import reactor.function.Function;
+
 /**
  * @author Spencer Gibb
  */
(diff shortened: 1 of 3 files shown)


spring-cloud-consul-sample/src/main/java/org/springframework/cloud/consul/sample/SampleApplication.java
```diff
@@ -82,7 +84,9 @@ public ServiceInstance me() {
 
 	@RequestMapping("/")
 	public ServiceInstance lb() {
-		return loadBalancer.choose(appName);
+		Function<String, ServiceInstance> f = loadBalancer::choose;
```
spencergibb: Using a method reference :-)

spencergibb:
[Here](https://github.com/orfjackal/retrolambda#backported-language-features) are the supported language features.

philwebb:
Looks interesting, but I'd approach with caution personally. I guess it depends how much benefit you get from the switch. I wonder if debugging is affected?

dsyer:
Having gone through this with spring boot metrics, I would say it's worth the ugliness in our code for the guarantee that it definitely works for all our Java 7 users. We can add some Java 8 native samples - that might address the legitimate demands of the Java 8 constituency.

PR #44, opened 2015-06-18
Change consul api call in ConsulServerList to get health info as well + new filter

nmarasoiu:


nmarasoiu:
this PR would benefit much from J8 lambdas, and HOFs like map for prop extractions.

The prev version had a bug: consul does not has a way to get info such as health check status by service instance id; but only by service name.

The higher level view, the aim is to have a selection of the "acceptable" service instances.
This means the non-critical status ones, perhaps further filtered (if we have an actual list after filtering).
Each service instance has 2categories of status: serf status and health check status.
The health checks is further divided into service and host (node) ones.
Each category (service, node) has multiple or no checks defined.
The node checks are indirectly obtainable from consul (there is api to get all checks running on node, and if we filter them by serviceId="", what remains, if any, are the node level checks, e.g. memory one).
My idea is to combine (aggregate) all health checks statuses by min.
From the instances, we select top instances by status.What do you think?

spring-cloud-consul-discovery/src/test/java/org/springframework/cloud/consul/discovery/ServiceCheckServerListFilterTests.java
```diff
@@ -0,0 +1,216 @@
...
+
+	@Test
+	public void singleInstanceAllGood() {
+		NewService serviceInstance = createInstance();
+		register(serviceInstance);
+		assertEquals(1, listOfServers(filter).size());
+	}
+
+    @Test
+    public void singleInstanceWarn() {
```
spencergibb: Mixed spaces and tabs here

spencergibb:
I think it's pretty complex.  I'd like the option to have just the service checks alone without the node checks.  Maybe it could be opt-in.

nmarasoiu:
I will think tommorow of the options. Maybe the default filter will be a simpler class, and we will also offer a complex scenario such as this, which seems common sense when you think of all the cases and implications, and also offer that to be wired up if they want. I can make this complex class more configurable, create 1 or 2 more simple ones, and reuse code between them.

spencergibb:
That would make me feel better about it.  Thanks for the continued work you are doing!

nmarasoiu:
So the most simple solution is
1. Filtering only by service checks
2. All services must have checks defined

In that case, passing a query parameter named "passing" to the service/checks endpoint yields only service instances with all statuses green.

We can do this in either of 2 places:
1. In a filter - and fully replacing the instances list with a new one coming from a consul rpc
2. As a configurable behavior of ConsulServerList: meaning on a boolean, switching from getting catalog services to getting services' with checks all green (which is a different api call to consul)

Also, an ugly part is that Ecwid does not provide a parameter for this, so that would mean appending "?passing" to our serviceName parameter, and rely on the fact that ecwid appends the parameter text mode to the URL.

nmarasoiu:
@spencergibb what do you think? 
In fact, what we seem to need is pluggable instances list providers, not just filters.
Than we would have 3 such implementations:
1. the current one (extracted): gets all service instances from catalog.
2. the one for this PR: gets all service instances which have at least 1 associated health check, and which have all the associated checks green.
3. picks up the highest ranking instances (ranked by health level); where health is a min-aggregation of the checks for the service instance; so this returns the same as 2 if at least one is fully green, or returns the warn ones if no green found.

than this may be a filter:
4. picks up the highest ranking instances (ranked by node' health level); where node health is a min-aggregation of the checks for the node (those with serviceId="" on the /node/checks)

spencergibb:
I think there is some confusion at to what parts of Ribbon do.  `ServerList` gets the list of servers.  `IPing` 'Interface that defines how we "ping" a server to check if its alive.' `ServerListFilter` 'allows for filtering the configured or dynamically obtained list of candidate servers with desirable characteristics.'  `ServerList` and `ServerListFilter` get called very often and I think it is the wrong place to put expensive API calls.  `ServerListFilter` could be suited for filtering by tags for example.  `IPing` is where to check the aliveness of a server.

nmarasoiu:
Ok, I had a few clarifications with Consul Armon:
- there is a consul api call to get all services with their checks, including service with no checks, and potentially filter them by all-green services; 
- that api also transparently adds the node' checks to the service checks

This means that for most scenarios, a single consul api call gets all the information needed on the client to determine the useful list.

As a consequence, we have 2 scenarios:
1. We want the green services. That means all services that do not have any checks below green. It will include services without a check. And will include the node checks for free, transparently.
2. We want the greens, but if no greens found, get the yellows. That means, give me the top service instances, but not critical ones.

For scenario 1, instead of the current strategy in ConsulServerList, we need to call another consul api, that will get the green services. A de-duplication on the client side may be necessary. That can be the default strategy.

For scenario 2, we need to call the same consul api as in scenario 1, but without the "?passing" query parameter. A heap based, or simpler client side checks aggregation + filtering strategy will be used.

As you can see, these are first-and-only consul calls. Using the filter mechanism for them is hacky. The filtering is still useful, for scenario 2, for various filtering strategies based on the statuses as already retrived from consul by ConsulServerList. 

Of course, to allow this, ConsulServerList needs to contain a new field, statusList, or lowestStatusOptional = an aggregate by min of the statuses for the service/node.

How I would do it, is, driven by a configuration, have the ConsulServerList make the service-with-health call with "?passing" or not, and when not, cache the status results in the ConsulServer objects, as list or as a single status.

nmarasoiu:
@spencergibb waiting for your feedback

spencergibb:
lots to digest, I'm just getting back to work after a week+ off.

spencergibb:
@nmarasoiu are you still interested in this?

ibzx:
I am not the original issue author but I am interested in this as it resolves my Issue reported https://github.com/spring-cloud/spring-cloud-consul/issues/116

spencergibb:
Closing for lack of feedback by the author and I don't think this is the right approach.

PR #45, opened 2015-06-18
publish servlet context as tag

nmarasoiu:


nmarasoiu:
for this to work full-vertical up to feign self-generating remoting proxy, I will try to see, other than adding meta with tags to consulServer, where else i need to propagate the tag to append to the balancing url, and I hope we will be able to iterate on that. bundling microservices in web application servers is probably a reasonable use case for deployment

spencergibb:
Do we want to have a check if the contextPath is empty or slash?  Seems like you would only publish that tag if it had meaningful data.

nmarasoiu:
Right, we can put the tag only if not empty and not "/"

nmarasoiu:
But that will imply if on both sides (publish/consume).
I find it more straight to always publish..

spencergibb:
Since consul is an outside entity, I would always check on the consuming side if the tag exists and the value is not empty.

nmarasoiu:
updated with condition to publish on non empty non slash

Issue #46, opened 2015-06-18
Throw an exception when sub folders are creating under the context


jmirc:
I am using the sample project and if I create the following structure the following exception is thrown

```
[...]

PR #47, opened 2015-06-19
Throw an exception when sub folders are creating under the context

jmirc:
Consul client returns nodes and leaves. This fix will save only the leaves in the properties map.

### DIFF
--- a/spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulPropertySource.java
+++ b/spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulPropertySource.java
@@ -22,12 +22,11 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
-import org.springframework.core.env.EnumerablePropertySource;
-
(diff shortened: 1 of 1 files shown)


jmirc:
Fix the bug #46

spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulPropertySource.java
```diff
@@ -55,9 +55,10 @@ public void init() {
 
 		if (values != null) {
 			for (GetValue getValue : values) {
-				String key = getValue.getKey().replace(context, "").replace('/', '.');
-				String value = getDecoded(getValue.getValue());
-				if (StringUtils.hasText(key)) {
+                String key = getValue.getKey();
+                if (!StringUtils.endsWithIgnoreCase(key, "/")) {
```
spencergibb: Please update formatting, tabs instead of spaces, thanks!

jmirc: done

PR #56, opened 2015-07-08
Fixes

nmarasoiu:
Hi,

I will leave the current company but I want to continue on these PRs until we merge them.

Additional to the other PR, here is a fix to the way time is computed on sending heartbeats, because I had windows of time when services where not green due to incorrect scheduling of heartbeats.


.travis.yml
```diff
@@ -1,22 +0,0 @@
-language: java
```
spencergibb: can't delete this file

spencergibb:
There are so many changes outside the scope of what you are proposing.

spencergibb:
maybe you can start a new PR with just the changes you want committed?

Issue #66, opened 2015-07-30
Support consul ACL tokens


bpgray:
It would be helpful to support ACL tokens for consul calls, for the use case where ACLs have been established in Consul and are required for kv store access and/or service registration.

If ACLs are in place, there are two ways to provide the token to consul (per consul.io) - either at the agent level (with the acl_token config parameter), or on a call specific basis. In some use cases, multiple applications/services will share an agent and may have different acl tokens defined. In these cases, setting the acl token at the agent level is not sufficient. By defining the token in a bootstrap.yml file, spring cloud consul could then provide the token for each call.

The Ecwid/consul-api supports providing a token in the KeyValueClient, but currently does not have the same support in the AgentClient. I'll open up an issue with consul-api to support tokens in the AgentClient, and I think that can be resolved by following the same pattern that exists in the KeyValueClient.

spencergibb:
Sounds good. Can you put a link to the `consul-api` issue when you create it?

bpgray:
Issue opened to support tokens in consul-api AgentClient
https://github.com/Ecwid/consul-api/issues/22

bpgray:
ACLs support has been added to the consul-api AgentClient in version 1.1.6.
Commit is here: https://github.com/Ecwid/consul-api/commit/5420831f46a9842218ad652508eff433b8e0f9b2

bpgray:
Ticket opened and pull request submitted to consul-api to support deregistering services with token
https://github.com/Ecwid/consul-api/issues/27

spencergibb:
Closed via 5154b91855fa1521dbd855e47bffcd1cb7c9b5f3

PR #80, opened 2015-09-04
add support for acl tokens in discovery and config

bpgray:
Pull request against https://github.com/spring-cloud/spring-cloud-consul/issues/66

This establishes two different acl tokens, one for registering the service in discovery and one for retrieving keys in config.

At that point Ecwid/consul-api does not support tokens for deregistering services. I'll open a new issue with consul-api to address.

Also, I was unable to build this due to some upstream dependencies (spring-integration-bom) not being available.

### DIFF
--- a/spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulConfigProperties.java
+++ b/spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulConfigProperties.java
@@ -37,4 +37,6 @@ public class ConsulConfigProperties {
 
 	@NotEmpty
 	private String profileSeparator = ",";
+
+	private String aclToken;
 }
--- a/spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulPropertySource.java
+++ b/spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulPropertySource.java
@@ -37,12 +37,14 @@ import com.ecwid.consul.v1.kv.model.GetValue;
 public class ConsulPropertySource extends EnumerablePropertySource<ConsulClient> {
 
 	private String context;
+	private String aclToken;
 
 	private Map<String, String> properties = new LinkedHashMap<>();
 
-	public ConsulPropertySource(String context, ConsulClient source) {
+	public ConsulPropertySource(String context, ConsulClient source, String aclToken) {
 		super(context, source);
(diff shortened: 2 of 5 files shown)


spencergibb:
So if it doesn't support deregistering with an acl, will this need to wait until `consul-api` does support it?

bpgray:
I think this probably needs to wait. Having register without deregister isn't entirely helpful.
If you like the approach, you could merge the k/v functionality into config and hold off on the discovery functionality.
Here is the ticket and pull request in consul-api: https://github.com/Ecwid/consul-api/issues/27
Once that goes, I can update this pull request to support deregister as well.

bpgray:
@spencergibb turns out we don't need tokens to deregister services, so this should be good to go. I also took some time to test both the discovery and config token functionality, and all checked out.

I put some detail on the situation with tokens and the deregister endpoint over on consul-api if you're interested: https://github.com/Ecwid/consul-api/pull/28

Also, let me know if you would like some tests for the acl token functionality, and I can work on a pull request next week.

Thank you!

spencergibb:
Sounds good. If you get tests by teusday, great. Otherwise I'll merge and you can add tests later in the week.

bpgray:
I think there are a couple of options for testing the ACL functionality.

If you want to go a functional route, I setup an example that secures consul with ACLs, registers a service using an allowed ACL, and then verifies the service was registered. Here is the branch: https://github.com/bpgray/spring-cloud-consul/tree/test-discovery-acl

Note, that I wasn't able to get this working in Travis CI (it couldn't find the config directory), which I think has something to do with the username/permissions in the .travis.yml. The tests did work for me locally.

This example still needs a bit of polish, including needing to add anonymous permissions to register the other test services (currently they are all ACL rejected, although the tests still pass).

Another approach would be to use mocks in the testing. It looks like mockito is being using in spring-cloud-config. Let me know if you would like to go that route, and what dependency to use.

I probably won't be able to develop the testing further until later in the week, so maybe best to merge this pull request as is.

bpgray:
Functional test now work on my fork and branch with travis at https://github.com/bpgray/spring-cloud-consul/tree/test-discovery-acl

I had to make a couple minor adjustments to .travis.yml to get this working on my fork. Let me know if you like this approach, and I can send a pull request. (I'll revert these adjustments in the pull request so it works in your repo).

spencergibb:
@bpgray go ahead with a new PR for the integration tests.

PR #82, opened 2015-09-08
Add tests for discovery ACL functionality

bpgray:
This PR secures the consul test instance with a default ACL policy of 'deny', and then runs a functional service registration test with a given ACL.

For cleanliness, the secured consul instance should allow registrations for the other tests without an ACL. Other tests do currently pass, so this isn't necessarily required given the current test design.

### DIFF
--- a/src/main/bash/local_run_consul.sh
+++ b/src/main/bash/local_run_consul.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
 consul agent -server -bootstrap-expect 1 -data-dir /tmp/consul -ui-dir `dirname $0`/../../../src/test/resources/consul_ui -config-dir `dirname $0`/../../test/resources/consul_config &
 # wait for consul to elect a leader before sending acl
-sleep 5000
+sleep 5
 curl -X PUT -d @`dirname $0`/../../test/resources/consul_acl/consul_discovery_client_acl.json http://localhost:8500/v1/acl/create?token=2ee647bd-bd69-4118-9f34-b9a6e9e60746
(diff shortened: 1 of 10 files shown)


bpgray:
Suspect build is failing due to #83
The build worked on my fork in travis after commenting out the ./docs/src/main/asciidoc/ghpages.sh command

src/main/bash/local_run_consul.sh
```diff
@@ -1,2 +1,5 @@
 #!/bin/bash
-consul agent -server -bootstrap-expect 1 -data-dir /tmp/consul -ui-dir `dirname $0`/../../../src/test/resources/consul_ui
+consul agent -server -bootstrap-expect 1 -data-dir /tmp/consul -ui-dir `dirname $0`/../../../src/test/resources/consul_ui -config-dir `dirname $0`/../../test/resources/consul_config &
```
spencergibb: Can we create a 2nd bash script, maybe `travis_run_consul.sh`, that has the config for tokens? That way I can run without tokens normally, but for builds use the new script.

bpgray:
Failing due to #84 
A script in the PR is removed by ghpages.sh

src/main/bash/local_run_consul.sh
bpgray: 2nd bash script created and now used in .travis.yml

spencergibb:
I haven't forgot about this. I need to make it so these tests only run on the travis build, not on a local build.

bpgray:
I think the tests will still work on a local build providing you haven't hardened your local consul instance with different acl keys. The test basically makes sure the tokens works against a hardened consul instance.

spencergibb:
Thats the thing, I don't want to run it with acls.

spencergibb:
Squashed, merged with a bit of polish via 12af0f532e1b60c51bf3e6e64fc0daacffa197e6

spencergibb:
@bpgray this is merged, but the config tests fail now, so for now they are disabled.

Issue #101, opened 2015-10-20
Address IP is ignored when registering service


nickcodefresh:
I'm  trying to run my service locally under Eclipse, but get it to register within Consul. I therefore need to provide a hostname to consul, so that clients use the service running on localhost. When Spring cloud consul registers the service the JSON passed to ```/v1/agent/service/register``` is:

```
{
    "ID": "mio-event-service:docker-compose:18102",
    "Name": "mio-event-service",
    "Tags": [],
    "Port": 18102,
    "Check": {
        "Interval": "10s",
        "TTL": "30s",
        "HTTP": "http://sss:18102/health"
    }
}
```

Note that the there is no ```Address``` element supplied. Looking at ```org.springframework.cloud.consul.discovery.ConsulLifecycle.register()``` I can't see it setting the address anywhere.

I can manually register the service via POSTMAN with an address, so I know it's supported:

```
{
    "ID": "1111",
    "Name": "mio-event-service2",
    "Tags": [],
    "Address": "localhost",
    "Port": 18102,
    "Check": {
        "Interval": "10s",
        "TTL": "30s",
        "HTTP": "http://169.254.133.254:18102/health"
    }
}
````

So my question is, how to tell Spring consul to supply an address in the JSON?

Thanks

spencergibb:
Yes, this is missing.

nickcodefresh:
Is this on the road map?  If not, I'll raise a pull request

spencergibb:
A pull request would be welcome.

aatarasoff:
Create pull request #105

PR #105, opened 2015-10-27
#101 add hostname from consul properties as service address for service and its management

aatarasoff:
test are included


spencergibb:
Merged with a little polish via d799efeae1878c981416a5a102933b2800037f80

PR #113, opened 2015-11-19
Added support to various formats in which configuration can be specified in consul

srikalyan:
Added the following changes
1. Added support to specify configuration in consul in different formats i.e.,
   configuration as key value pairs, configuration as property file contents

   In configuration as key value pairs format, users are allowed to specify configuration in consul
   using key value pairs where a value is either a constant or is a subkey

   In configuration as property file contents, users are allowed to specify configuration in consul
   with a single key and value as the contents of a properties file. (each line is separted using
   new line and any line starting with # is ignored

2. Made sure the original key value style of specifying configuration works and still is the default.

3. Made sure that Properties style works if config is overridden.

### DIFF
--- a/spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulConfigProperties.java
+++ b/spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulConfigProperties.java
@@ -42,8 +42,8 @@ public class ConsulConfigProperties {
 	private String consulConfigFormat = ConsulConfigFormat.KEY_VALUE.name();
 
 	/**
-	 * If consulConfigFormat is ConsulConfigFormat.PROPERTIES then the following field is used as key
-	 * to look up consul for configuration.
+	 * If consulConfigFormat is ConsulConfigFormat.PROPERTIES or ConsulConfigFormat.YAML
+	 * then the following field is used as key to look up consul for configuration.
 	 */
 	@NotEmpty
 	private String consulConfigPropertiesKey = "properties";
--- a/spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulConfigProperties.java
+++ b/spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulConfigProperties.java
@@ -46,7 +46,7 @@ public class ConsulConfigProperties {
 	 * then the following field is used as key to look up consul for configuration.
 	 */
 	@NotEmpty
-	private String consulConfigPropertiesKey = "properties";
+	private String consulConfigDataKey = "data";
 
 	private String aclToken;
 }
--- a/spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulConfigProperties.java
+++ b/spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulConfigProperties.java
@@ -38,5 +38,16 @@ public class ConsulConfigProperties {
 	@NotEmpty
 	private String profileSeparator = ",";
 
+	@NotEmpty
+   private String consulConfigFormat = ConsulConfigFormat.KEY_VALUE.name();
+
+   /**
+    * If consulConfigFormat is ConsulConfigFormat.PROPERTIES then the following field is used as key
+    * to look up consul for configuration.
+    */
+	@NotEmpty
+   private String consulConfigPropertiesKey = "properties";
+
+
 	private String aclToken;
 }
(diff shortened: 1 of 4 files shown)


srikalyan:
Thank you for the awesome library. It made it so easy. We need the configuration to be specified as properties and I am hoping I will add json and yml support soon.

spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulConfigProperties.java
```diff
@@ -38,5 +38,16 @@
...
 
+	@NotEmpty
+   private String consulConfigFormat = ConsulConfigFormat.KEY_VALUE.name();
+
+   /**
+    * If consulConfigFormat is ConsulConfigFormat.PROPERTIES then the following field is used as key
+    * to look up consul for configuration.
+    */
+	@NotEmpty
+   private String consulConfigPropertiesKey = "properties";
```
srikalyan: I am using spaces don't know why this is bit off :(.

spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulConfigProperties.java
```diff
@@ -38,5 +38,16 @@
 	@NotEmpty
 	private String profileSeparator = ",";
 
+	@NotEmpty
+   private String consulConfigFormat = ConsulConfigFormat.KEY_VALUE.name();
```
srikalyan: If you intend to use tabs instead of spaces then I can fix this.

spencergibb: Yes, we use tabs.

spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulPropertySource.java
```diff
@@ -53,25 +55,91 @@ public ConsulPropertySource(String context, ConsulClient source, String aclToken
...
+      }
+
+      for (GetValue getValue : values) {
+         String key = getValue.getKey().replace(context, "");
+         if (!consulConfigProperties.getConsulConfigPropertiesKey().equals(key)) {
+            continue;
+         }
+         final String value = getDecoded(getValue.getValue());
+         // values should be key=value\nkey1=value....
+         final String[] propertyLines = value.split("\n");
```
spencergibb: I wonder if there is a facility in spring-framework or spring-boot that could parse it for you. Maybe @dsyer would know.

artembilan: `StringToPropertiesConverter`, or just:
```
Properties props = new Properties();
// Must use the ISO-8859-1 encoding because Properties.load(stream) expects it.
props.load(new ByteArrayInputStream(source.getBytes("ISO-8859-1")));
```
directly.
The `TestPropertySourceUtils.convertInlinedPropertiesToMap()` does the trick as well.

spencergibb: @artembilan nice!

@srikalyan also, looks like YAML could be handled easily with `YamlProcessor` or `YamlPropertiesFactoryBean` fairly easily (though that can come later).

spencergibb: @srikalyan `StringToPropertiesConverter` isn't public, so use the two lines @artembilan mentioned.

For YAML,
```
YamlPropertiesFactoryBean yaml = new YamlPropertiesFactoryBean();
yaml.setResources(new ByteArrayResource(value.getBytes()));
Properties props = yaml.getObject();
```

srikalyan:
@artembilan thank you for the advice. I've updated the logic using props.load.
@spencergibb I've updated the diff with tabs instead of spaces.

I will add json and YAML support soon but in this PR though. I hope that is ok.

srikalyan:
BTW, is there a release scheduled soon if so I will try to get json and YML bits soon.

srikalyan:
@spencergibb, I just saw your YAML comment, I will update it right now.

srikalyan:
@spencergibb thanks to your snippet, I've added YAML support too. Please let me know if you need JSON support as well.

spencergibb:
@srikalyan I'll get a more detailed look in a day or two. JSON isn't needed at this point.

srikalyan:
@spencergibb thank you for the help and quick responses. I will wait for the snapshot build.

spencergibb:
@srikalyan have you signed the [contributor's agreement](https://support.springsource.com/spring_committer_signup)? I'm ready to merge.

srikalyan:
@spencergibb, I just did it. Please let me know if you need anything else and thank you for your help :) and such awesome library :+1: .
Your confirmation number is 149820151120080006

spencergibb:
Squashed and merged with some polish, a test and some documentation via 60b98a6edfff796cdee7dadd7ae63f7d6db7326b

PR #120, opened 2015-12-11
Extendable discovery configuration

### LINKED ISSUE #119: Cannot customize ConsulLifecycle
I noticed two problems when trying to use a subclass of `ConsulLifecyce`:
- in multiple configuration classes injection of dependencies is not really injecting them, but creating a separate copy (because it's attempted via a method call on the same class)
- there is no ConditionalOnMissingBean on ConsulLifecycle. Marking second implementation as `@Primary` starts both lifecycles and overriding beans by name is not the best way.

mabn:
Some fixes/refactoring in autoconfigurations which make providing own implementations of certain beans possible (in particular ConsulLifecycle)

Fixes #119

### DIFF
--- a/spring-cloud-consul-bus/src/main/java/org/springframework/cloud/consul/bus/ConsulBusAutoConfiguration.java
+++ b/spring-cloud-consul-bus/src/main/java/org/springframework/cloud/consul/bus/ConsulBusAutoConfiguration.java
@@ -23,6 +23,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.bus.BusAutoConfiguration;
 import org.springframework.cloud.bus.event.RemoteApplicationEvent;
+import org.springframework.cloud.consul.ConditionalOnConsulEnabled;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.integration.dsl.IntegrationFlow;
--- a/spring-cloud-consul-bus/src/main/java/org/springframework/cloud/consul/bus/ConsulBusAutoConfiguration.java
+++ b/spring-cloud-consul-bus/src/main/java/org/springframework/cloud/consul/bus/ConsulBusAutoConfiguration.java
@@ -23,7 +23,6 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.bus.BusAutoConfiguration;
 import org.springframework.cloud.bus.event.RemoteApplicationEvent;
-import org.springframework.cloud.consul.ConditionalOnConsulEnabled;
 import org.springframework.context.annotation.Bean;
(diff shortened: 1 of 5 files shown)


spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulConfigBootstrapConfiguration.java
```diff
@@ -34,21 +34,18 @@
 public class ConsulConfigBootstrapConfiguration {
 
 	@Configuration
-	@EnableConfigurationProperties
+	@EnableConfigurationProperties(ConsulConfigProperties.class)
```
spencergibb: I'd rather stick to the way we were creating the properties, please.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClientConfiguration.java
```diff
@@ -32,45 +33,42 @@
 @Configuration
 @ConditionalOnConsulEnabled
 @ConditionalOnProperty(value = "spring.cloud.consul.discovery.enabled", matchIfMissing = true)
-@EnableConfigurationProperties
+@EnableConfigurationProperties({LifecycleProperties.class, ConsulDiscoveryProperties.class, HeartbeatProperties.class})
 public class ConsulDiscoveryClientConfiguration {
 
 	@Autowired
 	private ConsulClient consulClient;
+	@Autowired
```
spencergibb: Don't autowire beans created in this class, pass them as parameters to the method.

spencergibb:
@mabn any response to my comments? If not, I'm going to close this.

mabn:
@spencergibb done

spring-cloud-consul-bus/src/main/java/org/springframework/cloud/consul/bus/ConsulBusAutoConfiguration.java
```diff
@@ -68,16 +69,17 @@ public ConsulOutboundEndpoint consulOutboundEndpoint() {
 
 	@Bean
 	public IntegrationFlow cloudBusConsulOutboundFlow(
-			@Qualifier("cloudBusOutboundChannel") MessageChannel cloudBusOutboundChannel) {
+			@Qualifier("cloudBusOutboundChannel") MessageChannel cloudBusOutboundChannel,
```
spencergibb: Why did you touch files outside of discovery?

mabn: As I mentioned in #119 
> in multiple configuration classes injection of dependencies is not really injecting them, but creating a separate copy (because it's attempted via a method call on the same class)

so i fixed all the occurrences I was able to find

spencergibb: Please limit your changes to discovery. This is out of scope for this PR.

spencergibb:
Squashed and merged with a little polish via ccea77d6615a22783daecbede15deb12dc6cef04

PR #122, opened 2015-12-14
Skips docker's interfaces

zanella:
The valid IP can be used for communicating w/ docker and w/ the outside, which is probably what you want when developing


dsyer:
Didn't this code live to spring-cloud-commons?

spencergibb:
@dsyer yes, but it looks like consul hasn't moved over yet.

zanella:
Port this one over there then? Assuming the skipping docker ifaces makes sense

spencergibb:
@zanella yes, let's port, but maybe more generically? Like a property that is a list of regex's to skip and some documentation? I don't like hardcoding things like "docker".

PR #126, opened 2016-01-08
Allow multiple dataKey so properties can be loaded from multiple keys

split3:
For Property and Yaml, it would be nice to support multiple dataKeys to allow various properties to be stored in different keys potentially.


split3:
Not sure why the checks failed, it appears like an issue with gradle dependency resolution

split3:
Seems the dependency failures are happening on the master as well, not sure what is needed to get done to fix that.  But are there any thoughts around this pull request?

PR #127, opened 2016-01-08
Allow service that will be defined to have the name customizable

split3:
We use ACLs to control what services can be registered based on various business units.  Since the consul registry doesn't support a hierarchy (similar to the KV store) we use a prefix on the service name, but use that prefix in the KV store as part of the hierarchy.

Allowing the service name to be customized allows this functionality to work.

### DIFF
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
@@ -74,6 +74,8 @@ public class ConsulDiscoveryProperties {
 
 	private int catalogServicesWatchTimeout = 2;
 
+	private String service;
+	
 	private String instanceId;
 
 	private String scheme = "http";
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulLifecycle.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulLifecycle.java
@@ -177,6 +177,12 @@ public class ConsulLifecycle extends AbstractDiscoveryLifecycle {
 	protected boolean isEnabled() {
 		return this.properties.getLifecycle().isEnabled();
 	}
+	
+	@Override
+	protected String getAppName() {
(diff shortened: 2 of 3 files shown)


split3:
@spencergibb - I've rebased which should fix the dependency resolution issues that were fixed in master.  Could you give this a look and let me know thoughts on it?

spencergibb:
You'll have to explain more and show me why this can't be handled with `spring.application.name`.

split3:
Again this is around ACLs and restrictions.  We namespace services with a business unit and then the service name.

`serviceName=abc-foo-service`

And we control access to the KV store and provide that Business Unit with access to the KV tree at

`/service/abc`

In which they would then access it via the `spring.application.name`

`/service/abc/abc-foo-service`

This is redundant and isn't as ideal.  Allowing the service name to be explicitly set if desired seems like a reasonable approach.

spencergibb:
You didn't show what `spring.application.name` is, I assume `foo-service`?

> In which they would then access it via the spring.application.name

> /service/abc/abc-foo-service

Do you mean ` /service/abc/foo-service`. If not, this still doesn't make sense to me.

split3:
Sorry if I wasn't completely clear, yes, `spring.application.name` is foo-service.  But we need the service when it is registered to have a different name `abc-$spring.application.name`.  That way when properties are looked up for this particular service as well it will use the path `/service/abc/foo-service` instead of `/service/abc/abc-foo-service`

spencergibb:
And this simplifies things because you only have to set the acl on `/service/abc` rather than for every `/service/{app}`?

split3:
Yes, we have dozens of business units and each business unit has several dozens of services.  So it greatly simplifies the ACL configurations along with keeping the single leaf node under `/service` from growing huge and unmanageable.

spencergibb:
OK, I belive I understand now.

spencergibb:
Merged with a slight change (`serviceId` vs `service`) via 2664b20f654405796c62da3d4fc2da7263ee32aa

split3:
Thanks for the merge, though I'd suggest that if you don't like `service` using `serviceName` instead as from a Consul nomenclature standpoint `instanceId` really should be called `serviceId` as that is what it is related to.  I used service to match the Consul API nomenclature.  So maybe it would make more sense as `serviceName` and possibly another pull request to refactor instanceId to serviceId.

PR #129, opened 2016-01-13
Add support for registering using an alternate external port

darylrobbins:
Sometimes the port that clients use to access a service is not the same port that the service is listening on. Maybe you have a web server sitting in front or maybe you're running in a docker container. In either case, one needs the ability to specify this alternate external port.

This PR adds an additional option to ConsulDiscoveryProperties for the external port. If it is provided, it is used when registering the service, otherwise it falls back to status quo behaviour.

### DIFF
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
@@ -63,6 +63,8 @@ public class ConsulDiscoveryProperties {
 
 	private String hostname;
 
+	private Integer externalPort;
+
 	private Lifecycle lifecycle = new Lifecycle();
 
 	/**
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
@@ -30,6 +30,8 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.cloud.util.InetUtils;
 
 /**
+ * Defines configuration for service discovery and registration.
+ *
  * @author Spencer Gibb
  */
 @ConfigurationProperties("spring.cloud.consul.discovery")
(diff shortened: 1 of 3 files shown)


spencergibb:
Just to clarify, you want the advertised port and the health check port to be different? Not sure that makes sense for the "web server in front" case, but may make sense for the docker case.

darylrobbins:
That's true for a web server with the agent running locally, it could really do the health check against either port. What I need is the ability for the port that is advertised to be different than the port that the application is listening on -- specifically for docker.

spencergibb:
How about just `port` instead of `externalPort`? Have you signed the [contributor's agreement](https://github.com/spring-cloud/spring-cloud-consul#sign-the-contributor-license-agreement).

darylrobbins:
I'm fine either way but I figured all the use of port might be confusing, and this is the specific port that is used to access the service externally. I have not. I can sign that now.

darylrobbins:
Contributor's agreement signed.

spencergibb:
In spring-cloud-netflix it is `nonSecurePort` or `securePort`. I'd prefer `port` with some javadoc's which show up when IDE's autocomplete properties. The other prop's need them too :disappointed:, but that's another story.

darylrobbins:
Alright, sounds good. I'll make the update. Is there anything special I need to do to make a release of the project for our own use, while we wait for a release incorporating this change?

I updated all the version numbers as you had from your last release but am still getting this error:

```
[ERROR]   The project org.springframework.cloud:spring-cloud-consul:1.0.0.ENTRUST (/Users/daryl/code/thirdparty/spring-cloud-consul/pom.xml) has 1 error
[ERROR]     Non-resolvable import POM: Could not transfer artifact org.springframework.cloud:spring-cloud-consul-dependencies:pom:1.0.0.ENTRUST from/to spring-snapshots (http://repo.spring.io/libs-snapshot-local): Failed to transfer file: http://repo.spring.io/libs-snapshot-local/org/springframework/cloud/spring-cloud-consul-dependencies/1.0.0.ENTRUST/spring-cloud-consul-dependencies-1.0.0.ENTRUST.pom. Return code is: 409 , ReasonPhrase:Conflict. @ line 60, column 16 -> [Help 2]
```

Any assistance would be greatly appreciated!

spencergibb:
That looks like maven can't resolve you're version, not sure I can be much help there.

darylrobbins:
Updates completed

spencergibb:
Nice! Thanks for the additional documentation.

spencergibb:
Squashed and closed with an additional test (the default port is non-zero) via 104131ba3506be7c1104f754b00a9437e336968d

PR #130, opened 2016-01-13
Allow multiple dataKey so properties can be loaded from multiple keys

split3:
For Property and Yaml formats it would be nice to be able to have multiple keys be used.

### DIFF
--- a/spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulConfigProperties.java
+++ b/spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulConfigProperties.java
@@ -48,7 +48,7 @@ public class ConsulConfigProperties {
 	 * then the following field is used as key to look up consul for configuration.
 	 */
 	@NotEmpty
-	private String dataKey = "data";
+	private String [] dataKey = {"data"};
 
 	private String aclToken;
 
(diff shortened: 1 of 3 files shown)


split3:
Sure, we use ACLs to control various access in consul KV store, so engineers might have access to some paths while DBAs might have access to others.  Having the ability to split the properties up accordingly like this and control that access is helpful.

split3:
For instance engineers might have access to the keys in **application.properties** but only DBAs might have access to **jdbc.properties**.

spencergibb:
You can do that without multiple data keys. What you're proposing is having different permissions for "directories" not leaf keys.

split3:
How exactly can you accomplish that without the multiple data keys.  Those are the keys used for lookup with YAML and PROPERTIES formats?

spencergibb:
Wouldn't `application.properties` and `jdbc.properties` already be separate keys?

split3:
Yes they are separate keys, but the way the PROPERTIES and YAML format works is it loads the contents of whatever is located at `data-key` thus just one key, unless I'm missing something.  So we aren't able to load 2 separate data-keys.

spencergibb:
The keys would look something like this
```
/config/<appname>,<profile>/data
/config/<appname>/data
/config/apppication,<profile>/data
/config/application/data
```
correct?

split3:
Yes but what we want to be able to do is have something more supported like

```
/config/<appname>,<profile>/application.properties
/config/<appname>,<profile>/jdbc.properties
/config/<appname>/application.properties
/config/<appname>/jdbc.properties
/config/apppication,<profile>/application.properties
/config/apppication,<profile>/jdbc.properties
/config/application/application.properties
/config/application/jdbc.properties
```

As engineers would have ACL access control to modify things within the application.properties and DBAs would have ACLs for jdbc.properties.

Obviously application.properties and jdbc.properties could be any values like `data, jdbc`, or whatever.  This isn't hugely critical so if it isn't seen as valuable we can work around it other ways.

spencergibb:
My take is that is what profile is for.

PR #131, opened 2016-01-18
Refactoring serviceId to serviceName

split3:
The field `serviceId` should really be called `serviceName` to match the consul nomenclature as `instanceId` is really the unique serviceId.


split3:
Based on the Consul Agent API:
https://www.consul.io/docs/agent/http/agent.html#agent_service_register

split3:
@spencergibb Any thoughts on this?

PR #138, opened 2016-02-06
Added support for customizing address for service

split3:
We are seeing an issue in which the wrong IP address is used when a machine has multiple ethernet adapters.

For instance we have the Consul Agent binding to eth0 on a particular machine, but when the service gets registered it is using the IP address from eth1 instead which causes problems as that isn't routable for the service.

Generally the address shouldn't be required as this isn't needed by Consul Service Definition by default.  If the address isn't provided then the agent's address will be used instead (this generally helps in the docker world as well).

The default behavior is set to now omit the address and use the agents IP, we can change this if desired by I think might be the most ideal approach in these situations and should work for almost all other.  The option for Lookup allow the existing behavior.

I didn't add to this pull request, as it might not be needed, but further enhancement might be added to get a named adapter instead of just find non-loopback found.

We might also want to pull some of this support down further into health checks as they can also be affected by the presence of multiple adapters.  I may add that to another pull request later on.

### DIFF
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
@@ -25,8 +25,8 @@ import java.util.Map;
 import lombok.AccessLevel;
 import lombok.Data;
 import lombok.Getter;
+import lombok.NoArgsConstructor;
 import lombok.Setter;
-import lombok.extern.apachecommons.CommonsLog;
 
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.cloud.util.InetUtils;
@@ -38,7 +38,7 @@ import org.springframework.cloud.util.InetUtils;
  */
 @ConfigurationProperties("spring.cloud.consul.discovery")
 @Data
-@CommonsLog
+@NoArgsConstructor(access=AccessLevel.PRIVATE)
 public class ConsulDiscoveryProperties {
 
 	protected static final String MANAGEMENT = "management";
@@ -86,7 +86,12 @@ public class ConsulDiscoveryProperties {
 	 * Use ip address rather than hostname during registration
 	 */
 	private boolean preferIpAddress = false;
-
+	
+	/**
+	 * Source of how we will determine the address to use
+	 */
+	private boolean preferAgentAddress = false;
+	
 	private int catalogServicesWatchDelay = 10;
 
 	private int catalogServicesWatchTimeout = 2;
@@ -108,8 +113,6 @@ public class ConsulDiscoveryProperties {
 	 * This allows filtering services by a single tag.
 	 */
 	private Map<String, String> serverListQueryTags = new HashMap<>();
-
-	private ConsulDiscoveryProperties() {}
 
 	public ConsulDiscoveryProperties(InetUtils inetUtils) {
 		this.hostInfo = inetUtils.findFirstNonLoopbackHostInfo();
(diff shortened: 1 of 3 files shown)


spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
```diff
@@ -86,7 +86,12 @@
 	 * Use ip address rather than hostname during registration
 	 */
 	private boolean preferIpAddress = false;
-
+	
+	/**
+	 * Source of how we will determine the address to use
+	 */
+	private AddressSource addressSource = AddressSource.AGENT;
```
spencergibb: I don't think an enum is needed, a boolean is fine.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulLifecycle.java
```diff
@@ -81,7 +82,9 @@ protected void register() {
 		Assert.notNull(service.getPort(), "service.port has not been set");
 		String appName = getAppName();
 		service.setId(getServiceId());
-		service.setAddress(properties.getHostname());
+		if(properties.getAddressSource() != AddressSource.AGENT) {
```
spencergibb: I'd rather not change the default behavior.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
split3: Yea enum probably isn't but did it that way for possibly future enhancement to support possibly named interface.  I can remove that if needed, figured an enum is pretty lightweight and simple.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulLifecycle.java
split3: If you'd rather not change the default that is fine, though I would imagine this would become a very common thing people would override (at least in our organization it would be every single project)

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
split3: I was thinking enum also because we could basically have the values like AGENT, IP, HOSTNAME at that point and possibly simplify the preferIpAddress also.  Would you rather me roll all that support up as well into the enum so the value is more apparent?

spencergibb:
I don't like mixing IP into this. I'd rather go with the boolean and not change the default. If enough users want it we can change it later.

split3:
I can pull is all back to just a basic boolean, I'm just not a fan of several booleans that participate in particular values as the understanding of what values override and have precedence becomes quickly blurred.  The way I see it the address has 3 possible value options

- empty (so it uses the agent address)
- hostname (so it uses an unresolved dns name or fqdn)
- ip (using an IP address of first non-loopback interface).

To me the enum makes sense here as I'm being very prescriptive about which option I expect to be used.  When my service definition is registered with the agent.

I have rebased with associated changes switching to a simple boolean and preserving the existing defaults.

spencergibb:
Thanks for the updates. 3 enums doesn't cover the 4 states of two booleans (t - t, t - f, f - t, f - f). I prefer the booleans for now.

split3:
Ok, fair enough, then we should be good here then with this.

spencergibb:
Yup, thanks for working with me on this. I appreciate your work and being amenable to change. I'll merge today sometime.

spencergibb:
Merged with a bit of polish via 21fe4b9ad3b3b3da522705658e0b629915bdd0c3. Thanks again!

PR #139, opened 2016-02-10
Add anonymous acl to fix config and discovery tests

bpgray:
This resolves the test issues created with #82

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -135,22 +135,5 @@
 				</plugins>
 			</build>
         </profile>
-		<profile>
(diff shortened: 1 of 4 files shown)


spencergibb:
Squashed and merged via 7c0a801986fbd983e679ff531bc194e8df1e8ea3. Turns out if you're not running with consul using ACL's, it doesn't matter if you set a token. Thanks @bpgray! This will be demoed tomorrow at our [Utah Spring User Group](http://www.meetup.com/utahspring/events/226893345/).

PR #140, opened 2016-02-11
Travis for forks

bpgray:
Sometimes it's helpful for forked repos to verify build functionality with travis prior to submitting pull requests.
This modification should leave the build functionality unaltered when preformed by the upstream repo (spring-cloud/spring-cloud-consul), but limit forked repos to just running tests (and skip deploy/install tasks).
Also, without this change forked repositories will fail with travis because they don't have permissions to complete the deploy task.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,9 @@ script:
 - if [ "${TRAVIS_REPO_SLUG}" == "spring-cloud/spring-cloud-consul" ]; then
     '[ "${TRAVIS_PULL_REQUEST}" != "false" ] || ./mvnw --settings .settings.xml deploy -nsu -Dmaven.test.redirectTestOutputToFile=true';
     '[ "${TRAVIS_PULL_REQUEST}" = "false" ]  || ./mvnw --settings .settings.xml install -nsu -Dmaven.test.redirectTestOutputToFile=true';
+  else
+    ./mvnw --settings .settings.xml test -nsu -Dmaven.test.redirectTestOutputToFile=true
   fi
-- ./mvnw --settings .settings.xml test -nsu -Dmaven.test.redirectTestOutputToFile=true
 env:
   global:
   - GIT_NAME="Spencer Gibb"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -15,8 +15,12 @@ install:
 - ./docs/src/main/asciidoc/ghpages.sh
 script:
 - ./src/test/bash/travis_run_consul.sh
-- '[ "${TRAVIS_PULL_REQUEST}" != "false" ] || ./mvnw --settings .settings.xml deploy -nsu -Dmaven.test.redirectTestOutputToFile=true'
-- '[ "${TRAVIS_PULL_REQUEST}" = "false" ]  || ./mvnw --settings .settings.xml install -nsu -Dmaven.test.redirectTestOutputToFile=true'
+- if [ "${TRAVIS_REPO_SLUG}" == "spring-cloud/spring-cloud-consul" ]; then
+    '[ "${TRAVIS_PULL_REQUEST}" != "false" ] || ./mvnw --settings .settings.xml deploy -nsu -Dmaven.test.redirectTestOutputToFile=true';
```
dsyer: Is this YAML valid? Don't we need a shell command to be a single element in a YAML array?

dsyer:
Thanks. We fixed it in a similar way independently (262ae1741f31cd), and for all the Spring Cloud projects.

.travis.yml
bpgray: It looks like the YAML spec considers this a flow scalar.
http://www.yaml.org/spec/1.2/spec.html#id2760844

2.18 provides an example of a multi-line flow scalar. The spec states.
"All flow scalars can span multiple lines; line breaks are always folded."

My forked build passes:
https://travis-ci.org/bpgray/spring-cloud-consul/builds/108651413
You can see how the multi-line flow scaler has the line breaks folded in the console output. I think this is a good indicator that the travis-ci parser is following the spec, and bash doesn't care much about line breaks for this command.

I thought about refactoring the ${TRAVIS_PULL_REQUEST} conditional into an if-then-else for cleaner readability, but tend towards preserving existing code in pull requests.

Let me know if you prefer another style, and I'm happy to refactor.

bpgray:
Thank you @dsyer. It looks like our comments crossed ... I like your solution.

Issue #154, opened 2016-03-15
Whitelisting/blacklisting services based on consul tag


ahawtho:
We expose a number of our services in a consul cluster.  Some of those services we want to expose through a Zuul proxy, but some are intended only to be used by internal services.  We could blacklist them in the proxy configuration, but that could potentially leave new services exposed unintentionally.  We could whitelist in the proxy configuration, but that requires a change to the proxy each time we want to release a new service.

My thought was that we could tag desired services with a specific value and filter services that don't match that value in the ConsulDiscoveryClient.  More generally, a "filter" in ConsulDiscoveryClient might be useful for other developers.  Is there something like this that I haven't discovered yet?

Thanks for any pointers or suggestions.

spencergibb:
Nothing like it. Unsure how you would use this in zuul? Another option would be Consul ACL's. Run the zuul server with an ACL that is restricted to see services that are "proxyable".

ahawtho:
We've configured Zuul to use the ConsulDiscoveryClient to discover its services rather than using Eureka, so the possibility of filtering that list makes this a pretty simple change (the filter concept could be used to filter on other criteria as well).

The reason I want to use consul tags is because our system allows consul tags to be specified within the service configuration.  Each team developing a service can simply denote whether their service is intended to be publicly visible simply by adding a tag in their config.  Using the ACL would make this less automatic since we'd still have to orchestrate adding each service to the ACL's definition.

I think I see how it could be done in a fairly simple manner.  I'll try posting a PoC in the next week or so.

PR #161, opened 2016-03-23
Add ConsulServiceDiscoveryFilter support

ahawtho:
- Add ConsulServiceDiscoveryFilter to filter services upon discovery
- Apply filter to ConsulDiscoveryClient while listing services
- Add two implementations: an AcceptAllDiscoveryFilter and a
  TagMatchingDiscoveryFilter
- Add configuration properties for the TagMatchingDiscoveryFilter's
  includes and excludes

Fixes gh-154

### DIFF
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClient.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClient.java
@@ -16,9 +16,18 @@
 
 package org.springframework.cloud.consul.discovery;
 
+import static org.springframework.cloud.consul.discovery.ConsulServerUtils.findHost;
+
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+
+import org.springframework.boot.autoconfigure.web.ServerProperties;
+import org.springframework.cloud.client.ServiceInstance;
+import org.springframework.cloud.client.discovery.DiscoveryClient;
+import org.springframework.cloud.consul.discovery.filters.ConsulServiceDiscoveryFilter;
+import org.springframework.util.StringUtils;
 
 import com.ecwid.consul.v1.ConsulClient;
 import com.ecwid.consul.v1.QueryParams;
@@ -28,15 +37,7 @@ import com.ecwid.consul.v1.agent.model.Self;
 import com.ecwid.consul.v1.agent.model.Service;
 import com.ecwid.consul.v1.health.model.HealthService;
 
-import org.springframework.boot.autoconfigure.web.ServerProperties;
-import org.springframework.cloud.client.DefaultServiceInstance;
-import org.springframework.cloud.client.ServiceInstance;
-import org.springframework.cloud.client.discovery.DiscoveryClient;
-import org.springframework.util.StringUtils;
-
 import lombok.extern.apachecommons.CommonsLog;
-
-import static org.springframework.cloud.consul.discovery.ConsulServerUtils.findHost;
 
 /**
  * @author Spencer Gibb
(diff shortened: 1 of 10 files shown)


spencergibb:
@ahawtho have you signed the [contributor's agreement](https://github.com/spring-cloud/spring-cloud-consul#sign-the-contributor-license-agreement)?

spencergibb:
I'm not sure the `ConsulDiscoveryClient` is the best place for this to go. Ribbon already has the concept of server list filters and I feel like that is more appropriate place. We could use our commons `ServiceInstance.getMetadata` to implement filters for all spring cloud discovery implementations (eureka, zookeeper and consul). Obviously #162 would need to be implemented for that to work. What do you think @ahawtho @dsyer @marcingrzejszczak?

dsyer:
Thumbs up for https://github.com/spring-cloud/spring-cloud-consul/issues/162

ahawtho:
@spencergibb yes, I signed it yesterday before submitting the PR.

I actually looked at <code>ServiceInstance.getMetadata</code> first, but tags are exposed as a list in a consul service, and the metadata is a Map, so there's some dissonance with the representation.

One approach might be to use each tag as a key with an empty or other sentinel string value.  The issue I saw with that is the potential for namespace collision with k-v metadata when consul publishes 1.0 .

Another approach is to use a single _tags key mapped to a delimited string, but I was hoping to avoid the potential escaping/parsing issues, since tags are free-form.

Neither of those were really compelling to me. Do you have a suggestion or preference for how to expose the tags via <code>ServiceInstance.getMetadata</code>?  If so, I can try to code that up instead.

Thanks...

spencergibb:
@ahawtho I'd prefer to go the `ServiceInstance.getMetadata` route. If a tag has an `=` split it into key/value for the map, otherwise put them in something like `_tags`. We'll worry about what consul metadata is in 1.0 when it drops.

ahawtho:
@spencergibb Ok, sounds good.  I'll give it a shot.

It turns out that changing the ```ConsulDiscoveryClient``` wasn't enough for my use-case anyway.  Spring Cloud Netflix configures a ```DiscoveryClientRouteLocator``` which uses the ```DiscoveryClient.getServices()``` to obtain a list of the service names (which ironically, I didn't change here).  After that, the ```ConsulRibbonClientConfiguration``` uses the ```ConsulServerList``` afterward to obtain the individual server instances.  It doesn't touch ```DiscoveryClient``` for finding instances at all.  I'm wondering if```ConsulServerList``` should be using the ```ConsulDiscoveryClient``` instead of duplicating that work itself?  It already has a dependency on ```ConsulDiscoveryProperties```, so it wouldn't be unreasonable.

```HealthServiceServerListFilter```/```ConsulServer``` could then be implemented in terms of ```ConsulServerInstance```, and if we added a ```@ConditionalOnMissingBean``` for ```ribbonServerListFilter()```, clients could override the ```ServerListFilter``` bean to achieve what I was hoping to do.

spencergibb:
Funny you should mention that. I'm thinking consul and zookeeper don't need their own server list impl, just a `DiscoveryClientServerList` or something.

ahawtho:
Ah, yeah... that would be even better.

sirlatrom:
@ahawtho Would it make sense to allow users of `ConsulDiscoveryClient` to specify a tag in an overloaded `getInstances()` to allow querying for service instances with the given tag as part of this PR, since it introduces Consul's tag concept in general?

spencergibb:
@ahawtho I'm going to go ahead and close this. Feel free to comment if you feel otherwise.

ahawtho:
@spencergibb - this is fine.  I think the <code>DiscoveryClientServerList</code> is a better way to go, with tags through the <code>ServiceInstance.getMetaData()</code>.  Thanks...

Btw, any effort underway to switch to a <code>DiscoveryClientServerList</code>?  Could that work also be used for a kubernetes implementation?  I saw the spring-cloud-kubernetes project, but is it possible there's some synergy here?  I think we will need something approaching this in the near future (~ 6 months), so I may have cycles to work on it.

spencergibb:
@ahawtho it's something I was planning on doing in the next few months for 'Camden' release train.

PR #164, opened 2016-04-01
Add the ability to pass query parameters for service discovery

jjathman:
For specific services our company only runs them in one of our data centers. To be able to discover the appropriate service we need to pass in a specific data center with the query to Consul. This PR adds the ability to pass a custom `QueryParams` object to the `getInstances` call. It didn't seem like this was possible using the existing Consul client. Added one simple test for this, it's not really easy to test extensively without being able to assume there are multiple DCs available.

This is my first PR to this project so please let me know if there's anything else I need to do.

I have signed the contributor's agreement.

### DIFF
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClient.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClient.java
@@ -40,6 +40,7 @@ import static org.springframework.cloud.consul.discovery.ConsulServerUtils.findH
 
 /**
  * @author Spencer Gibb
+ * @author Joe Athman
  */
 @CommonsLog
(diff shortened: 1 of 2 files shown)


spencergibb:
Interesting. Would it be useful to have a configuration property for the datacenter rather than (or in addition to) using `ConsulDiscoveryClient` explicitly?

jjathman:
Possibly, in our use case we need to decide this at runtime (normally we will just use the default DC but a few of our services only run in a remote DC) so we probably wouldn't use a configurable value but I can see the appeal. It doesn't seem as though Consul has a good solution to our problem so we are rolling our own using the K/V store and specifying the DC when we search for specific services.

spencergibb:

jjathman:
Yep I have.

spencergibb:
Merged with a little polish via bcad4ca683bc7cbc274654e97fb8bca45c42acd9

PR #166, opened 2016-04-02
add ability to configure a default service tag to query by

Fitzoh:
This is an enhancement along the lines of `serverListQueryTags`. Instead of having to define a query tag to filter each serviceId by, you can now add a default query tag that all service id's will be filtered by.

This also adds a convenience method to `ConsulDiscoveryProperties` to select the appropriate tag for a given serviceId, delegating to the default tag if a corresponding tag is not found in the `serverListQueryTags` map.

### DIFF
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
@@ -115,6 +115,11 @@ public class ConsulDiscoveryProperties {
 	private Map<String, String> serverListQueryTags = new HashMap<>();
 
 	/**
+	 * Tag to query for in service list if one is not listed in serverListQueryTags.
+	 */
+	private String defaultQueryTag;
+
+	/**
 	 * Add the 'passing` parameter to /v1/health/service/serviceName.
 	 * This pushes health check passing to the server.
 	 */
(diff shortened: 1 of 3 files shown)


Fitzoh:
Oops, just noticed the java 8 test failure, updated.

spencergibb:
@Fitzoh Can you give me a little more concrete example of what this solves?

Fitzoh:
Sure @spencergibb .  Our consul instance is currently shared across multiple environments, and we have 1 instance of each service on each environment.

Our route definitions look something like this:
```
zuul:
   routes:
    service1:
      path: /service1/foo/**
      serviceId: service1
    service2:
      path: /service2/bar/**
      serviceId: service2
    service3: 
      path: /service3/baz/**
      serviceId: service3
```

We only want services to talk to their associated services in the same environment, so we're using the server-list-query-tags to restrict each service lookup to the environment tag.
```
spring:
  cloud:
    consul:
      discovery:
        enabled: true
        server-list-query-tags:
          service1: ${environment}
          service2: ${environment}
          service3: ${environment}
```

This gets a little tedious as the number of services grow, so this PR adds in the ability to set up a default query tag for each environment.  The updated config would look like this:

```
spring:
  cloud:
    consul:
      discovery:
        enabled: true
        defaultQueryTag: ${environment}
```

While still giving you the ability to override the query tags for specific services if necessary.


```
spring:
  cloud:
    consul:
      discovery:
        enabled: true
        defaultQueryTag: ${environment}
        server-list-query-tags:
          service1: bar
```

spencergibb:

Fitzoh:
I think so, can't find a confirmation though. Just signed a new one, confirmation is 181320160603092133.

spencergibb:
Thanks @Fitzoh for the quick response.

spencergibb:
Closed with some polish via 2b38a3c2758e6b7972a68ddd0e70f4f1eb8ce73e

Issue #109, opened 2015-11-11
Dynamic Tags for service registration


mishadoff:
Is there a way to set `spring.cloud.consul.discovery.tags` programmatically based on some application logic?

My use case is following: I have some `bootstrap.jar` application which runs with `-Dloader.path` for additional jars as dynamic components. Each dynamic component has a specific interface, annotation `@Component` and a name. I want to expose all component names as a tags during service registration.

spencergibb:
No, there's not.

mishadoff:
It is a pity. 
Do you plan to add support for this at some point? Or any workarounds?

spencergibb:
You just opened the issue :-)

spencergibb:
Pull requests are welcome.

venilnoronha:
As discussed on #168, this behavior can be achieved by extending `ConsulLifecycle` and overriding the `register(NewService)` method.

PR #168, opened 2016-04-04
Enabled Consul service customization

venilnoronha:
I've created a callback interface called `ConsulServiceCustomizer`. Several implementations of this callback can be wired to `ConsulLifecycle` during bean creation. `ConsulLifecycle` would execute these callbacks as the last step in its `register()` method. `ConsulServiceCustomizer` can be used to add additional tags / configure other parameters of the service. See #109.

Please review and pull. My CLA number is 149720151120072904.

Thanks,
Venil Noronha

### DIFF
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulLifecycle.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulLifecycle.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2015 the original author or authors.
+ * Copyright 2013-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulLifecycle.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulLifecycle.java
@@ -16,11 +16,14 @@
 
 package org.springframework.cloud.consul.discovery;
 
+import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
 
 import javax.servlet.ServletContext;
 
+import lombok.Getter;
+import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
 
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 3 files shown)


spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulLifecycle.java
```diff
@@ -106,6 +112,12 @@ protected void register() {
 			service.setCheck(createCheck(checkPort));
 		}
 
+		if (!this.serviceCustomizers.isEmpty()) {
```
spencergibb: Unneeded?

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulServiceCustomizer.java
```diff
@@ -0,0 +1,35 @@
...
+ * @author Venil Noronha
+ */
+public interface ConsulServiceCustomizer {
+
+	/**
+	 * Customize the Consul service.
+	 * 
+	 * @param service the {@link NewService} to customize.
+	 */
+	public void customize(NewService service);
```
spencergibb: This worries me a bit: exposing a library class as a supported api.

spencergibb:
You could also extend `ConsulLifecycle` and override `register(NewService)`.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulLifecycle.java
venilnoronha: Fine.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulServiceCustomizer.java
venilnoronha: I felt that the user case is similar to https://github.com/spring-projects/spring-boot/blob/master/spring-boot/src/main/java/org/springframework/boot/context/embedded/undertow/UndertowDeploymentInfoCustomizer.java. However, I get your point.

venilnoronha:
You mean to say, the original issue _(add additional tags)_ can be solved by overriding `register(NewService)`? If we don't need another API for customization, then we can close this PR.

spencergibb:
That's what I was suggesting. If we can solve the same problem by keeping the surface area the same, I'd prefer that. Is that satisfactory to you?

venilnoronha:
Cool.

Issue #181, opened 2016-05-09
High log volume when consul agent down


bpgray:
Sometimes consul agents fail, or need to be restarted for various reasons.

When the agent goes down, ConfigWatch.watchConfigKeyValues() will start logging errors that include the full stack trace. If it's down for a couple of minutes, this can result in a several megabyte log file.

To prevent massive log files, there are some options:
1) Implement a longer fixedDelayString, which helps but also has side effects.
2) Don't log the full stack trace in the catch block
3) Use a back off pattern on error. One simple implementation would be for the class to track errors and introduce a sleep if the last few calls resulted in an error.

If there is a preferred option, I can work on a pull request.

spencergibb:
1) is a user option, so nothing to do there.
2) is the simplest. I'd log the full error on debug, msg on error. I'd start here.
3) We already use spring retry

PR #182, opened 2016-05-09
reduce logging on ConfigWatch connection failure

bpgray:
Implementation of option 2 for Issue https://github.com/spring-cloud/spring-cloud-consul/issues/181


bpgray:
@spencergibb With this pull request I modified the implementation of failFast in ConfigWatch to fail on the first query, but not if the agent is lost while running (after the first query). Before the failFast flag was used to determine the logging approach, which may not have been the intended functionality. Another approach would be to eliminate the use of failFast in ConfigWatch, as I think we will already fail during initialization in ConsulConfigProperties if the agent is down.

spencergibb:
Merged with some polish via c6cdaa13626e41e8c870f1b51eacb09f0958317b

PR #190, opened 2016-05-24
Fixes #189 - Improve Consul TTL Check management

### LINKED ISSUE #189: Check TTL - Consul log missed TTL, is now critical
I experienced frequent consul log like that after reducing ttlValue param to 15s 
`2016/05/19 18:30:29 [WARN] agent: Check 'service:test-service-8090' missed TTL, is now critical`

After some code review and experimentation I see that we have the same behavior with default parameters (ttlValue = 30sec).
There is 3 parameters that are involved
* spring.cloud.consul.discovery.heartbeat.**ttlValue** _(default 30s)_ : Used to register TTL check on consul
* spring.cloud.consul.discovery.heartbeat.**intervalRatio** _(default 2.0 / 3.0)_ : Used to evalutate a rate at which we will send a ping to Consul
* spring.cloud.consul.discovery.heartbeat.**fixedRate** _(default 15s)_ : Used by [TtlScheduler](../blob/master/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/TtlScheduler.java) to check if we have to send a ping to Consul.

So if we put this all together (by default)
|-**t 0s** Our service is registered with a TTL and Consul wait a ping for 30s
|-**t 15s** TtlScheduler awake and check if heartbeat interval is past due (30 * 2 / 3 ~= 20sec) => **No PING**
|-**t 30s** Consul detect that our service TTL expire and log : **[WARN] agent:.... missed TTL, is now critical**
|-**t 30s** TtlScheduler awake and check if heartbeat interval is past due (and it is by 10sec) => **Send a PING**
|-**t 30s** Consul acknowledge our ping and log : **[INFO] agent: Synced check ...**
...etc.

This can only work if 
**fixedRate** > (**ttlValue** * **intervalRatio**) => ex : 21s with default values
OR
**fixedRate** <  (**ttlValue** * **intervalRatio** / **2**) => ex : 9s with default values

Maybe we could first set **fixedRate** to 21s by default and log a warning if fixedRate doesn't match **ttlValue** **intervalRatio** requirement.

A cleaner solution is to find a way to set **fixedRate** dynamically or to a very low value (1sec?) to avoid these side effect.

stephaneLeroy:
Remove spring.cloud.consul.discovery.heartbeat.fixedRate property and use a TaskScheduler (with FixedRateTrigger) to send Consul checks

### DIFF
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/TtlScheduler.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/TtlScheduler.java
@@ -17,24 +17,31 @@
 package org.springframework.cloud.consul.discovery;
 
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.*;
 
 import lombok.extern.slf4j.Slf4j;
 
 import org.joda.time.DateTime;
+import org.springframework.core.task.TaskExecutor;
+import org.springframework.scheduling.TaskScheduler;
 import org.springframework.scheduling.annotation.Scheduled;
 
 import com.ecwid.consul.v1.ConsulClient;
 import com.ecwid.consul.v1.agent.model.NewService;
+import org.springframework.scheduling.concurrent.ConcurrentTaskScheduler;
+import org.springframework.scheduling.config.Task;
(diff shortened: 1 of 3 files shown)


spencergibb:
Merged with some polish via cf998fffae4f3b90cd0209f0383efc29a9b0a25e

PR #193, opened 2016-06-01
Fix example segment of code

imduffy15:
The discovery client example code attempts to call a `getInstances` function on a non existing client, this should be `discoveryClient`


spencergibb:

imduffy15:
Just after doing so.

PR #195, opened 2016-06-07
Added support for customized configuration of the Management Port

aantono:
Added support for customized configuration of the Management Port. (useful when running separate mgmt context inside docker, with 2 port mappings)

(Contribution Agreement has been signed)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ target/
 _site/
 .idea
 *.iml
+*.ipr
 .factorypath
 *.swp
 /consul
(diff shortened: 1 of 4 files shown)


spencergibb:
In all my tests setting `management.port` is sufficient. Are you seeing otherwise?

aantono:
Setting `management.port` will actually make the Servlet Container Context Connector run on that port.  Which does not always match to the Docker port exposing map, when the app is running as a docker container.

For example, all our apps are deployed as Docker containers, so inside Docker container, the app service is running on port `8080` and the `management.port` is configured to be `8081`.  When ECS or Mesos launches the container, the external host mappings are never `8080` and `8081`, but are randomly selected ports.  We then internally in Docker launch the app with setting env variables to `SPRING_CLOUD_CONSUL_DISCOVERY_HOST`, `SPRING_CLOUD_CONSUL_DISCOVERY_PORT` and would like the ability to be able to configure what the management port should be when registered to Consul (since it is different than what it is inside the container).  So having `SPRING_CLOUD_CONSUL_DISCOVERY_MANAGEMENT_PORT` property would solve this problem.

Let me know if that makes sense, hopefully I've explained it coherently. ;)

spencergibb:
Closed with a little polish via 0fe6039cac5bdb8489f646ed9944e68f2fb6be37

Issue #210, opened 2016-08-18
Adds ability to deregister a service based on critical check state longer than a timeout


panga:
Based on this PR (https://github.com/hashicorp/consul/pull/2276) that will be released with Consul 7.0, please provide property to define and deregister zombie services automatically.

spencergibb:
That property will need to be added the consul library we use https://github.com/Ecwid/consul-api/issues

dwelch2344:
Was just about to file an issue about adding support for this. Will open an issue there and submit a PR, then follow up here. 

Assuming the property is there, where would be the best place to make it configurable? I'd imagine a property from `application.yml` would be desirable, no? 

Happy to contribute on this

spencergibb:
The issue is already there https://github.com/Ecwid/consul-api/issues/64.  I would assume the property would go in [`ConsulDiscoveryProperties`](https://github.com/spring-cloud/spring-cloud-consul/blob/master/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java)

spencergibb:
Ping @dwelch2344

PR #239, opened 2016-10-25
Adds property to deregister services in critical longer than timeout

donnabelldmello:
This PR adds a config property `healthCheckCriticalTimeout` which enables deregistering services that are in critical for longer than timeout. I have updated the consul API version to `1.2.1`.
(Refer to #210 for more information)

Thanks,
Donnabell

### DIFF
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2015 the original author or authors.
+ * Copyright 2013-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulLifecycle.java
```diff
@@ -136,6 +137,7 @@ protected void register() {
 		}
 		check.setInterval(properties.getHealthCheckInterval());
 		check.setTimeout(properties.getHealthCheckTimeout());
+		check.setDeregisterCriticalServiceAfter(properties.getHealthCheckCriticalTimeout());
```
spencergibb: Should only set of not empty (since there is no default).

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
```diff
@@ -70,6 +71,9 @@
 	/** Timeout for health check (e.g. 10s) */
 	private String healthCheckTimeout;
 
+	/** Timeout to deregister services critical for longer than timeout. */
```
spencergibb: Provide a format example like healthCheckTimeout above.

spencergibb:
ping @donnabelldmello on changes. Also needs to be rebased.

spencergibb:
Closed with some polish and test via 93948854012712aaee79e9330290274fa0d3aea7

PR #277, opened 2017-02-03
remove unnecessary reference to outer class

kamilszymanski:


spencergibb:
How about a test that shows invalid properties throwing an error?

kamilszymanski:
This PR was a false start since validation (within bootstrap context) happens faster than I thought and therefore removing Spring Boot 1.5 warning messages about missing validation for validation constraints should probably be handled in other place (module).

All that's left from this false start is a simple polishing of inner class references.

Issue #231, opened 2016-09-30
ConfigWatch does not pick up newly put config files in FILES mode


jfiedler:
Once I fixed #229 I ran into another inconsistency with ConfigWatch. The problem is that the internal Consul index handling in ConfigWatch prevents a newly stored config file to lead to a refresh event. Only an additional change to the config file is being picked up correctly.

Steps to repeat:
- Start server ind 'dev' profile with 'application.yml' file in Consul K/V store
- Put a new 'application-dev.yml' file in Consul (settings for server running dev profile)
  -> No refresh event (even though there are new settings that require a refresh)
- Change the 'application-dev.yml' config file that you just put there
  -> Refresh event is being fired

The root cause IMHO here is that ConfigWatch does not distinguish between the case that Consul has not been called yet (i.e. at startup time) and the case that a certain context path returned a 404 (because there was no file). The transition from 'I know there was no file' to 'there now is a file where there previously was none' should lead to a refresh event.

This problem is not specific for FILES config format. It should also be a problem for the other modes. However, in those modes the watch looks at the directory which is typically there (with new properties being added below that). For FILES mode, it is much more frequent that a newly watched element is being added.

spencergibb:
It's an interesting quandry, the 404/empty check was added to reduce churn (ie too many refreshes that weren't part related). Suggestions?

jfiedler:
I think that the whole watch handling in kind of odd in case of FILES mode. In an ideal world, you would only generate one refresh event if **any** of the watched context path changes. If I am not mistaken it would currently send one refresh per changed file (which is also churn). 'Any' context path above should include those where Consul does not have a file. This is necessary since deleting a file in Consul is also a config change. 

My suggestion therefore is to have a loop over the context paths and hash them along with the Consul index returned for them (assuming 0 in case of HTTP 404). The state of the watch would be this hash (not the individual Consul index numbers). If the hash is different after the loop you would issue a refresh.

I was planning to try this out with tests etc locally but did not have time to get to it. Therefore no new pull request yet ... :-(

dahamr:
Sorry, I missed this scenario in my PR to reduce churn. We are running into this as well. Our situation is because many services have no configuration in consul initially. Adding the consul folder and item does not produce a refresh event the first time. The second update does.

As for suggestions, I've been testing seeding the consulIndex with a value other than -1L to indicate ConfigWatch has been called at least once. The initial test worked but I'll try to be more thorough this time.

[...]

PR #296, opened 2017-04-10
ConfigWatch not issuing refreshEvent on new context

dahamr:
The refreshEvent() is not sent when the context is created after
the service has started. This is a simple change to indicate
that watchConfigKeyValues() has been called once even if the
context doesn't exist.

intended to fix gh-231


spencergibb:
From https://github.com/spring-cloud/spring-cloud-consul/issues/231#issuecomment-293014725

Added the simplest PR I could come up with. The intent is to support these scenarios now:
- first time call on an existing context, don't send refreshEvent() because its too early
- any call on a missing context should not call refreshEvent()
- if context is added after startup, call refreshEvent()
- Otherwise, if index changes, call refreshEvent()

Since release is soon, I was hoping to give you something in time so I don't have to come up with an alternative solution in house.

Issue #278, opened 2017-02-08
EnvironmentChangeEvent fires before config property values are updated on first event


joshfix:
I would like to execute code when a particular config value is modified in Consul. However, the very first time a config value is modified in Consul and the EventListener catches the EnvironmentChangeEvent, Spring seems to not have populated that config property value yet. Every time thereafter, the value is properly populated when the EnvironmentChangeEvent listener is invoked.

I have provided a sample application that replicates the behavior here:

https://github.com/joshfix/demo

spencergibb:
See [ConfigurationPropertiesRebinderAutoConfiguration.java#L75-L81](https://github.com/spring-cloud/spring-cloud-commons/blob/master/spring-cloud-context/src/main/java/org/springframework/cloud/autoconfigure/ConfigurationPropertiesRebinderAutoConfiguration.java#L75-L81)

> // After all beans are initialized send a pre-emptive EnvironmentChangeEvent
		// so that anything that needs to rebind gets a chance now (especially for
		// beans in the parent context)

Seems @dsyer is doing it on purpose.

joshfix:
I'm not sure that describes the issue I'm seeing.  My EventListener doesn't even seem to catch that initial empty event that you linked.  The issue I'm having happens after the first property value change in consul.  The very first time you change a value, the change event fires, but the property isn't updated yet.  I have to manually call RefreshEndpoint.refresh() for the updated value to be set.  Every change in consul thereafter works as expected -- the change event fires and the updated property value is set.

beeva-pablovaliente:
I am experiencing the same issue as mentioned #288 

My scenario is as follows:

* My application starts when consul is still empty
* After application is started, a key is written in consul (manually for example)
* The application cannot detect this first change until a new change is made again in consul (or refresh is called). When the second change happens, the event that arise, contains both updated keys.

Hope this helps.

Regards.

joshfix:
FWIW, @beeva-pablovaliente, I believe it is the same issue.  The only difference between your scenario and mine is I had an initial key/value set for the property I was fetching in my event listener.

spencergibb:
See https://github.com/spring-cloud/spring-cloud-consul/pull/297

spencergibb:
BTW, the updated test [failed](https://jenkins.spring.io/view/Spring%20Cloud/view/CI/job/spring-cloud-consul-master-ci/2300/) with version 0.7.2 of consul in CI. Updating to 0.7.5 let the test pass.

PR #297, opened 2017-04-10
Get seed indexes from consul in property source locator.

spencergibb:
These initial indexes can then be passsed to the `ConfigWatch` so that
it doesn't have to have an odd skip the first time.

fixes gh-231
fixes gh-278


spencergibb:
Closed via 5520dd2ba63090a3b83ec3a6f3ff5cb9bd4ac4a2

Issue #304, opened 2017-04-28
Add support for being able to set additional Consul tags via Env and System Properties


aantono:
Currently, if a `spring.cloud.consul.discovery.tags` property is set inside `bootstrap.yml` or inside `application.yml` files, it gets completely overridden by setting additional tags via command-line argument or Env variable, like setting `SPRING_CLOUD_CONSUL_DISCOVERY_TAGS` externally.  This behavior breaks the intent of being able to `augment` the list of tags, not replacing them completely.

I've tried doing something like `SPRING_CLOUD_CONSUL_DISCOVERY_TAGS="\${spring.cloud.consul.discovery.tags},external1=blah"`, but that has resulted in an infinite resolution loop [`Circular placeholder reference 'spring.cloud.consul.discovery.tags' in property definitions`] (rightfully so).

It would be great to be able to (somehow) augment the Consul Tags from external configs to add to the internally defined ones.  (Maybe doing something "clever" like asking the environment for all the property sources and iterating over them in their configured precedence order from bottom to top, collecting all the tags for `spring.cloud.consul.discovery.tags` key and adding them all together)

PR #305, opened 2017-04-29
Add support for creating composite list of consul tags for spring.cloud.consul.discovery.tags property

aantono:
Proposed implementation for #304

### DIFF
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulTagsEnvironmentPostProcessor.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulTagsEnvironmentPostProcessor.java
@@ -1,0 +1,26 @@
+package org.springframework.cloud.consul.discovery;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.env.EnvironmentPostProcessor;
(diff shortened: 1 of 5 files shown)


spencergibb:
Complicated but interesting. That's not a default behavoir for lists like this in all of spring boot/spring cloud. You could have a default tags and combine the two properties that would be much simpler.

aantono:
I've thought about the separate property names route, but had an issue with having to propagate custom configuration to all the applications having to merge the two.  The roll-up composite seemed like a more elegant solution, though I agree that everything else in Spring world for property sources is an override and not a composite.  I could see how there might be a reluctance to break that paradigm, so totally open to other solutions that would solve this use-case without having to make changes to every app's config. :)

lowzj:
I think, It may be simpler that if we add another `Map<String, String>` property `ConsulDiscoveryProperties#externalTags`, and convert it to `List<String>` and merge it into `ConsulDiscoveryProperties#tags`. Then we can add tags as below:

```
spring.cloud.consul.discovery.externalTags.test1=value1
spring.cloud.consul.discovery.externalTags.test2=value2
```

But this is not friendly to adding a non-KeyValue tag.

spencergibb:
We've decided as a team to not go against the default paradigm. Feel free to come up with another solution. Thanks for your time. We really appreciate the community involvement.

Issue #314, opened 2017-06-05
Allow optional use of servlet


vigozhang:
After upgraded spring-cloud-consul-discovery from 1.1.4.RELEASE to 1.2.0.RELEASE,non-web spring boot application occurs error during booting:

```
WARN 64141 --- [           main] s.c.a.AnnotationConfigApplicationContext : Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'org.springframework.cloud.client.serviceregistry.ServiceRegistryAutoConfiguration$ServiceRegistryEndpointConfiguration': Unsatisfied dependency expressed through field 'registration'; nested exception is org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'consulRegistration' defined in class path resource [org/springframework/cloud/consul/serviceregistry/ConsulAutoServiceRegistrationAutoConfiguration.class]: Unsatisfied dependency expressed through method 'consulRegistration' parameter 2; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'javax.servlet.ServletContext' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}

Parameter 2 of method consulRegistration in org.springframework.cloud.consul.serviceregistry.ConsulAutoServiceRegistrationAutoConfiguration required a bean of type 'javax.servlet.ServletContext' that could not be found.
```

urvish-everbridge:
Is there a work around for this issue ?

spencergibb:
No.

pwielgolaski:
I simply use servlet API, no read servlet container.
Only problem that I have with this it trigger web mode in spring boot application that I need to disable after that.

PR #372, opened 2017-11-04
Extract servlet logic to be included conditionally

pwielgolaski:
When classpath has ServletContext, include in spring context ConsulRegistrationCustomizer that could be used to customize consul tags

fixes gh-314

### DIFF
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/serviceregistry/ConsulRegistrationCustomizer.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/serviceregistry/ConsulRegistrationCustomizer.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright 2013-2016 the original author or authors.
+ *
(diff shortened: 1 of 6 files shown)


spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/serviceregistry/ConsulRegistrationCustomizer.java
```diff
@@ -0,0 +1,26 @@
...
+
+package org.springframework.cloud.consul.serviceregistry;
+
+import java.util.List;
+
+/**
+ * @author Piotr Wielgolaski
+ */
+public interface ConsulRegistrationCustomizer {
+	void customizeTags(List<String> tags);
```
pwielgolaski: FYI, I would be happy to consider here NewService, but wonder if we want to leak internals here.

spencergibb:
I fixed this in a different direction. Thanks for the contrivution.

Closed via d9b55bc

spencergibb:
And I changed my mind :-)

Issue #367, opened 2017-10-24
Can we have timer metrics while watchConfigKeyValues in ConfigWatch?


JagmohanSharma:
We need to keep track of consul calls being made for each consulIndexes so I think it would be a good idea to keep timer metrics around watchConfigKeyValues() to trace calls time taken.

I have made changes regarding this. Please let me know if this really adds value here.

spencergibb:
I think it probably would make sense, though the implementation will be different in 2.0.x with micrometer.

JagmohanSharma:
Yes, actually we are making use of com.ryantenney.metrics's metrics-spring provided annotation support for  Dropwizard Metrics library. which is not being used spring cloud consul. I will see if we can do this by existing libraries.

PR #382, opened 2018-01-02
Adding timer metrics for consul config watch

JagmohanSharma:
Integrating timer metrics for consul config watch using Netflix Servo library.

https://github.com/spring-cloud/spring-cloud-consul/issues/367


spencergibb:
We shouldn't be using servo, but the new micrometer metrics

JagmohanSharma:
I think we have java version differences in both the library. As micrometer is using java 1.8. Are we looking to make spring cloud consul to java 1.8 default?

spencergibb:
Master is Java 8

pom.xml
```diff
@@ -146,6 +147,12 @@
 				<artifactId>joda-time</artifactId>
 				<version>${joda-time.version}</version>
 			</dependency>
+			<!-- Micrometer dependency -->
+			<dependency>
+				<groupId>io.micrometer</groupId>
+				<artifactId>micrometer-spring-legacy</artifactId>
```
spencergibb: I'm a bit confused why you are pulling in the legacy support for spring boot 1.5.x. Spring Boot 2 already managed micrometer 1.0.x

JagmohanSharma: Updated pull request by removing legacy support dependency.

Issue #308, opened 2017-05-06
Support empty prefix


rnemykin:
Hi, i store in consul key smth like this
```
image.color.red=RED
image.color.green=GREEN
```
So many applications put such key/value to consul (A,B,C)and i want watch for them (for kyes with prefix _image.color.*_) from another application (D).
Currently i have to store properies like
```image/color/green``` because _org.springframework.cloud.consul.config.ConsulPropertySourceLocator_ works with '/' separator.

```
List<String> suffixes = new ArrayList<>();
if (this.properties.getFormat() != FILES) {
	suffixes.add("/");
} else {
	suffixes.add(".yml");
	suffixes.add(".yaml");
	suffixes.add(".properties");
}

String defaultContext = prefix + "/" + this.properties.getDefaultContext();
for (String suffix : suffixes) {
	this.contexts.add(defaultContext + suffix);
}
for (String suffix : suffixes) {
	addProfiles(this.contexts, defaultContext, profiles, suffix);
}
```

Also i would like to configure only key prefix 
`spring.cloud.consul.config.prefix=image.color`

and don't set properties like defaultContext, but it's _@NotEmpty_ and i can't set it to null (default is "application" which also brings inconvenience).
Similarly situation with _name_ property of _org.springframework.cloud.consul.config.ConsulConfigProperties_

May be intersect with
https://github.com/spring-cloud/spring-cloud-consul/issues/303

spencergibb:
Not sure this make sense. The property sources are already watched and a spring event `EnvironmentChangeEvent` is published with the keys that have changed.

rnemykin:
it makes some discomfort, if I specify 
`spring:
  cloud:
    consul:
      config:
        prefix: image.color`
, I expect that changing these keys (keys with such prefix) will update be handled, but it isn't.
Actually, with such config, spring do watching for keys
`[name='image.color/{spring.application.name}/'], ConsulPropertySource [name='image.color/application/']`

Because contexts for key watching building with some defaults values (defaultContext, name) and also separates with `/`. 

Until I looked in the source code I do not understand how can I set the configuration to get it working,
In docs (http://cloud.spring.io/spring-cloud-static/Dalston.RELEASE/#spring-cloud-consul-config-watch) nothing is said about it.

I would like the flexibility to configure such things

rnemykin:
up

spencergibb:
The prefix is a 'directory' or path in the consul heirarchy.

[...]

PR #392, opened 2018-02-06
Optional Prefix

pjhampton:
Makes a prefix optional.
If prefix is empty, then it will fall back on `${spring.application.name}`

Closes: https://github.com/spring-cloud/spring-cloud-consul/issues/308


spencergibb:
Some test failures

pjhampton:
@spencergibb They seem to be permission based. Can you please trigger a build? Seems there were two passing builds before this.

spencergibb:
Merged with some polish via 201375af297e357def7489e979124454936eb6e5

Issue #377, opened 2017-11-20
add @ConditionalOnMissingBean on ConsulConfigProperties in ConsulConfigBootstrapConfiguration


perdonare:
now I found It's difficult to customerized ConsulConfigProperties

spencergibb:
Can you explain a bit more about what you need?

perdonare:
Typically I have two sets of  config for consul  where confirm useful for our team and don't need two much config  ,and by set the flag I can choose  which one we can use , this need dynamic change the consul config , also   I want change the config key  from spring.cloud.consul.xxx  to   xxx.consul.xxx

spencergibb:
You can do that with Spring profiles. Why do you need a different prefix?

perdonare:
I want integration cloud consul to our  framework ,  and  no need too much care about spring cloud

spencergibb:
That doesn't make much sense to me

perdonare:
any way , it's convenient for other people use

spencergibb:
I don't know. We could have used the class name in `@EnableConfigurationProperties` instead of `@Bean`. You'd be in the same situation as now having to use the bean name. We'll never officially support using a different namespace.

spencergibb:
I'd like to see if others find it convenient and their reasons

edvinerikson:
I have a use case where I want to build a thin layer/framework around cloud config and provide a default context and prefix... The overall goal of the wrapper is to provide the software contracts that are required to run a service on our in-house platform such as service discovery and configuration management.

Instead of having all our services define where to look for default properties I would like to have the wrapper set it by providing a `ConsulConfigProperties` Bean. This would make it easier for the developers to fulfil the contracts that we have within our platform.

I currently did a dirty hack to get a POC working without a conditional bean.

```java
@AutoConfigureBefore(ConsulPropertySourceLocator.class)
public class CloudConfigAutoConfig {
    @Autowired
    private ConsulConfigProperties consulConfigProperties;

    @PostConstruct
    public void init() {
        consulConfigProperties.setPrefix("platform-config");
        consulConfigProperties.setDefaultContext("defaults");
    }

}
```

spencergibb:
PRs welcome

edvinerikson:
Awesome @spencergibb! I'll have a look! Though, be aware this will be my first contribution ever to Spring and a JVM open-source project.. :)

PR #422, opened 2018-05-09
Make ConsulConfigProperties bean conditional

edvinerikson:
I read the contribution guidelines and did the best I could! If you find something that you think I should change, feel free to mention it :)

Feedback would be awesome

Fixes GH-377

cc @spencergibb


edvinerikson:
I found another case where we also need a conditional.
https://github.com/spring-cloud/spring-cloud-consul/blob/master/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClientConfiguration.java#L64

Would you be okay with me adding a conditional there as well? If so, can I do it in this PR?

I found this one too in the bootstrap context but we don't currently need to configure in that context as of now (Might change as we are building this out): https://github.com/spring-cloud/spring-cloud-consul/blob/master/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/configclient/ConsulDiscoveryClientConfigServiceBootstrapConfiguration.java#L42

Issue #443, opened 2018-08-27
custom instanceID id not used for registration of management service which leads to inconsistency


jackhammer2k:
We're using a different port for the management stuff in our Spring Boot/Cloud application which leads to 2 services registered in Consul (real one and "management"). In addition we specify a custom instance ID and noticed now that that custom ID is not used for the "management" service. That's a bit confusing when looking into Consul because both services do not look like having any relationship to each other, which isn't true.

org.springframework.cloud.consul.serviceregistry.ConsulAutoRegistration:266
```
public static String getManagementServiceId(ConsulDiscoveryProperties properties, ApplicationContext context) {
		return normalizeForDns(IdUtils.getDefaultInstanceId(context.getEnvironment(), false)) + SEPARATOR + properties.getManagementSuffix();
	}
```
should be
```
public static String getManagementServiceId(ConsulDiscoveryProperties properties, ApplicationContext context) {
		return normalizeForDns(getInstanceId(properties, context)) + SEPARATOR + properties.getManagementSuffix();
	}
```

lomeshpatel:
@spencergibb Do you know if this was done on purpose? I can't see any reason why custom instanceId and serviceName shouldn't be applied to management service. I may very well be missing something here.

Please let me know, I have most of it coded and can have a PR up soon.

spencergibb:
PRs welcome.

PR #451, opened 2018-09-19
Apply customized instanceId and serviceName to management service

lomeshpatel:
- If instanceId and/or serviceName is customized, they're also applied
to the management service registration
- Updated documentation to provide information about management service

Fixes gh-443


spencergibb:
if the value returned from the consul api is null, I don't want to change it.

Issue #452, opened 2018-09-19
ConsulDiscoveryClient getServices() does not take default-query-tag into account


fuleow:
We have a mix of different types of services registered in Consul and have tagged our Spring Boot based services with a special tag (e.g. `sb`). We need a way to be able to filter the list of services returned by `ConsulDiscoveryClient` to only include services based on this tag.

Our use case is very similar to what's described in #225 but the difference is we do not know the service names so we cannot directly use the enhancements from that issue.

It would be nice if the `getServices()` method filtered the list of services based on the value of `default-query-tag` if it is set.

fuleow:
> @fuleow Hi, did you manage to find a solution to this issue?

Hi, please see the referenced PR for the solution. If you do not want to run your own patched version you can implement similar logic in a custom `DiscoveryClient`

mario45211:
+1

PR #453, opened 2018-09-19
Modify the getServices() method so that it filters on default-query-tag

fuleow:
This PR modifies `ConsulDiscoveryClient` so that its `getServices()` method takes into account the value of `default-query-tag` if it is set.

Further details are available in issue #452.

### DIFF
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClient.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClient.java
@@ -19,6 +19,7 @@ package org.springframework.cloud.consul.discovery;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import com.ecwid.consul.v1.ConsulClient;
 import com.ecwid.consul.v1.QueryParams;
@@ -119,12 +120,24 @@ public class ConsulDiscoveryClient implements DiscoveryClient {
 	public List<String> getServices() {
 		String aclToken = properties.getAclToken();
 
+		Map<String, List<String>> services;
 		if (StringUtils.hasText(aclToken)) {
-			return new ArrayList<>(client.getCatalogServices(QueryParams.DEFAULT, aclToken).getValue()
-					.keySet());
+			services = client.getCatalogServices(QueryParams.DEFAULT, aclToken).getValue();
 		} else {
-			return new ArrayList<>(client.getCatalogServices(QueryParams.DEFAULT).getValue()
-					.keySet());
+			services = client.getCatalogServices(QueryParams.DEFAULT).getValue();
+		}
+
+		return services.entrySet().stream()
+				.filter(e -> defaultQueryTagFilter(e.getValue()))
+				.map(Map.Entry::getKey)
+				.collect(Collectors.toList());
+	}
+
+	private boolean defaultQueryTagFilter(List<String> tags) {
+		if (StringUtils.isEmpty(properties.getDefaultQueryTag())) {
+			return true;
+		} else {
+			return tags.contains(properties.getDefaultQueryTag());
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@fuleow Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-consul&pullRequestId=453)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-consul&pullRequestId=453) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClient.java
```diff
@@ -119,12 +120,24 @@ private void addInstancesToList(List<ServiceInstance> instances, String serviceI
 	public List<String> getServices() {
 		String aclToken = properties.getAclToken();
 
+		Map<String, List<String>> services;
 		if (StringUtils.hasText(aclToken)) {
-			return new ArrayList<>(client.getCatalogServices(QueryParams.DEFAULT, aclToken).getValue()
-					.keySet());
+			services = client.getCatalogServices(QueryParams.DEFAULT, aclToken).getValue();
```
spencergibb: These methods should use the `getCatalogServices()` method that takes a tag parameter (which can be null, which is the default value for `properties.defaultQueryTag`). Otherwise, for a large deployment this would be very inefficient.

fuleow: Thanks for taking a look at this PR. I don't see a `getCatalogService()` method that can be used for this. The only ones which take a tag also take a service. The problem this is trying to solve is to query all services with a given tag, so the service name is not known upfront.

spencergibb: I see them. Maybe you need to rebase from master to get the latest version of consul-api.

fuleow: I see 2 methods which accept the `tag` parameter, and both require a `serviceName` to be provided
```
public Response<List<com.ecwid.consul.v1.catalog.model.CatalogService>> getCatalogService(String serviceName, String tag, QueryParams queryParams)
public Response<List<com.ecwid.consul.v1.catalog.model.CatalogService>> getCatalogService(String serviceName, String tag, QueryParams queryParams, String token)
```
I've looked at the [implementation](https://github.com/Ecwid/consul-api/blob/master/src/main/java/com/ecwid/consul/v1/catalog/CatalogConsulClient.java#L141) and `getCatalogServices(QueryParams queryParams)` used here makes a call to Consul's [list services' endpoint](https://www.consul.io/api/catalog.html#list-services) which does *not* accept a list of tags. The list services endpoint returns the entire catalog along with tags in a single request regardless of how large the deployment is.

The methods that accept a tag for filtering calls Consul's [list node for services' endpoint](https://www.consul.io/api/catalog.html#list-nodes-for-service) which does take in a tag but requires a specific service to be provided. [(implementation)](https://github.com/Ecwid/consul-api/blob/master/src/main/java/com/ecwid/consul/v1/catalog/CatalogConsulClient.java#L183)

spencergibb: Thanks, you are right. I guess our current implementation is no less performant :-)

fuleow: Great, I'm relieved to hear that. I've only started working with Consul recently and was worried I missed something.

spencergibb:
Since it is a change in functionality or behavior, it requires the CLA

pivotal-issuemaster:
@fuleow Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-consul&pullRequestId=453)!

varnson:
In Finchley, this method is useless, I don't know why catalogservice was called frequently.

PR #456, opened 2018-09-29
Added the instance id to the `DefaultServiceInstance`

TYsewyn:
Should be merged after https://github.com/spring-cloud/spring-cloud-commons/pull/419


ryanjbaxter:
@TYsewyn can you check this PR?  The build fails in CircleCi as well as on my machine

TYsewyn:
Hi @ryanjbaxter, I just ran the tests locally and everything is running fine.
The build on CircleCI failed because spring-cloud/spring-cloud-commons#419 wasn't merged into master back then.
First thing that comes to mind is that you don't have a consul instance running on localhost:8500.
Can you check again? Thanks!

Issue #254, opened 2016-11-18
Override of spring consul config not working?


evalsinca:
I have properties in my consul key value store that are loaded properly by my spring boot app.  However, overriding a Consul property by specifying a value on the command line does not seem to work.  Here are the details:  My consul data source contains a property "a.b.c" with value "foo".   Spring boot reads this value correctly.  However, when I run my app like so:

`java -jar api.jar --a.b.c="bar"`

I am expecting to see the value "bar". Instead, I still see "foo".  

Notice that if the property "x.y.z" is set in bootstrap.yml to "hello", and overridden on the command line like so:

`java -jar api.jar --x.y.z="world"`

the value of "x.y.z" is correctly reported as "world".   

Is property overriding supported for properties in Consul key value store?

spencergibb:
The consul config is given the highest priority by default.  You can set `spring.cloud.config.override-none=true` and it will make external config the lowest priority. You can set `spring.cloud.config.override-system-properties=false` which will put them below system properties, which will also make cli args higher priority.

patrickbray:
Thanks @spencergibb our team spent a while trying to figure out how to get this working before stumbling on this issue. Would you accept a PR to add the above to the docs? I would also add that these properties need to be set in Consul. I had originally tried setting them in `bootstrap.properties`

PR #497, opened 2019-02-20
gh-254: Add documentation on config overrides

patrickbray:
Ran into this issue again https://github.com/spring-cloud/spring-cloud-consul/issues/254 so figured I would submit a quick documentation PR.


spencergibb:
I'm a bit confused how config client properties `spring.cloud.config.override-none` have anything to do with consul config?

ryanjbaxter:
I think this is already clarified in this section of the docs

http://cloud.spring.io/spring-cloud-static/Greenwich.RELEASE/single/spring-cloud.html#overriding-bootstrap-properties

Is that not saying the same thing?

patrickbray:
> I think this is already clarified in this section of the docs
> 
> http://cloud.spring.io/spring-cloud-static/Greenwich.RELEASE/single/spring-cloud.html#overriding-bootstrap-properties
> 
> Is that not saying the same thing?

@ryanjbaxter yes definitely perhaps a link back would help here? I will close this PR now.

PR #471, opened 2018-11-12
Service check TTL

devpreview:
Closes #470

PR #564, opened 2019-06-12
fix #471 again-Service check TTL

varnson:
Though #471 had been closed, I think that the fix was not enough.
The real reason is, when the service was registered, the TTL was not set.
So we need to set TTL when register the service while RegisterHealthCheck() is false and heartbeatProperties.isEnabled() is true.
The settings are:
spring.cloud.consul.discovery.register-health-check=false
spring.cloud.consul.discovery.heartbeat.enabled=true

I had fixed it with my SpringCloud Finchley, and copied the changes to the master branch.
And I find that function createCheck() had been fixed with moving HealthCheckCriticalTimeout to the top of that function.

But there are still two places needed to be fixed!


spencergibb:
Why do you keep closing and opening new PRs?

varnson:
There are ci errors so I had to resubmit it. I'm sorry for it.

spencergibb:
No you don't. Pushing a new change should trigger CI and we can to. Resubmitting is worse.

PR #500, opened 2019-02-28
Fix deregister services critical not working when using TTL

NikoTung:
This PR add deregister services critical  timeout to service registry if the `healthCheckCriticalTimeout` parameter is present regardless of  health check type(http or ttl). 

fixes gh-490

Issue #490, opened 2019-01-29
After enable TTL, DeregisterCriticalServiceAfter not working .


NikoTung:
I used the following config to enable the TTL and DeregisterCriticalServiceAfter.

    spring:
       cloud:
         consul:
            host: localhost
            port: 8500
            discovery:
              instance-id: ${spring.application.name}:${vcap.application.instance_id:${spring.application.instance_id:${random.value}}}
             health-check-critical-timeout: 3m
             heartbeat:
                enabled: true
                ttl-value: 30

When the application crash(exit without executing  the deregister ),the application will be in the critical status on consul ui and never be removed by consul.

I read the code and found that the other parameters are ignored when the ttl are enable.

    public static NewService.Check createCheck(Integer port, HeartbeatProperties ttlConfig,
										 ConsulDiscoveryProperties properties) {
		NewService.Check check = new NewService.Check();
		if (ttlConfig.isEnabled()) {
			check.setTtl(ttlConfig.getTtl());
			return check;
		}
		...
		return check;
	}
So my question is: Why the DeregisterCriticalServiceAfter is ignored. 

Since it is not conflict with the ttl. And I check the [consul.io ](https://www.consul.io/api/agent/service.html)document.

I alse test with the following curl script.The application:test-critical will be removed by consul automatically after 1 minute

    curl -X "PUT" "http://localhost:8500/v1/agent/service/register" \
     -H 'Content-Type: application/json' \
     -d $'{
        "ID": "test-critical-9e7433ac909aaf67ce256818b8e14e28",
        "Address": "10.0.11.179",
        "Name": "test-critical",
        "Check": {
          "TTL": "20s",
          "DeregisterCriticalServiceAfter": "1m"
        },
        "Port": 8080
      }'

spencergibb:
I think that the `consul-api` library used to only support one `Check`, that is not the case anymore.

NikoTung:
I know the `consul-api` library used to only support one `Check`. I think that the `DeregisterCriticalServiceAfter ` is kind of consul agent behavior, and no matter what kind  of `check`  we specified ,the `DeregisterCriticalServiceAfter` should not be ignored when it is config.

scrat98:
Same issue. Could you please add DeregisterCriticalServiceAfter feature on TTL checks. Thanks!

spencergibb:
@scrat98 PRs welcome

PR #565, opened 2019-06-12
refix #471 for service-ttl error

varnson:
Though #471 had been closed, I think the fix was not enough.
The real reason is, when the service was registered, the TTL was not set.
So we need to set TTL when RegisterHealthCheck() is false and heartbeatProperties.isEnabled() is true.
The settings are:
spring.cloud.consul.discovery.register-health-check=false
spring.cloud.consul.discovery.heartbeat.enabled=true

I had fixed it with my SpringCloud Finchley, and copied the changes to master branch.
And I find that function createCheck() had been fixed with moving HealthCheckCriticalTimeout to the top of this function.

There are still two places need to be fixed!

### DIFF
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/serviceregistry/ConsulAutoRegistration.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/serviceregistry/ConsulAutoRegistration.java
@@ -128,6 +128,9 @@ public class ConsulAutoRegistration extends ConsulRegistration {
 			Assert.notNull(checkPort, "checkPort may not be null");
 			service.setCheck(createCheck(checkPort, heartbeatProperties, properties));
 		}
+		else if (heartbeatProperties.isEnabled() && service.getCheck() == null) {
+			service.setCheck(createCheck(0, heartbeatProperties, properties));
+		}
 	}
 
 	public static ConsulAutoRegistration managementRegistration(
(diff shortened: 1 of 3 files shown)


spencergibb:
Can you explain better the situation why you think this is a problem?

Also, tests are required.

varnson:
@spencergibb Consul supports two types of healthcheck, one is consul call the healthcheck interface， which is enabled when spring.cloud.consul.discovery.register-health-check=true, another is the service reports its own status to consul,which is enabled by spring.cloud.consul.discovery.heartbeat.enabled=true. If the consul server does not receive the status report from service in TTL time, the service will be marked as unavailable. 
In the method createCheck(), it will add TTL time to service registeration info when ttlConfig.isEnabled(), and health check interface will not be set. But when register-health-check is not true, createCheck() will not be called, and the TTL will not be set. It's the reason for #471. So we need to set TTL when heartbeat.enabled is true.
Are you means  heartbeat.enabled=true must  be set both with register-health-check=true? I think it's two types of healthcheck and heartbeat shoud not depend on register-health-check.
I will add test later.

varnson:
I found that the method  createCheck() was fixed by PR #500, and it was related to issue #490.
In #490, the user had set heartbeat.enabled to true, but had not set register-health-check to false.
But I don't think heartbeat.enabled must be used with register-health-check=true.

varnson:
I had added tests in my branch.

PR #492, opened 2019-02-11
Add instanceIdIncludeHostname config property

tianzhipeng-git:
Add instanceIdIncludeHostname config property to control 'whether hostname is included into the default instance id when registering service' #487

PR #570, opened 2019-06-29
fix #492-add hostname in instanceid

varnson:
#492 had been closed, but I need it, so I refixed it based on master branch.
Thanks to  @tianzhipeng-git.


eacdy:
HOPE this PR be merged soon.I need it, too. 
Or else I need to set `spring.cloud.consul.discovery.instance-id` or extend  `org.springframework.cloud.consul.serviceregistry.ConsulServiceRegistry#register` and reset instanceId.
Neither are good approach. 
I like this PR better. It's the best way I think. And I nearly submit a same PR.

PR #589, opened 2019-09-19
Added reactive support for service discovery

TYsewyn:


spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/reactive/ConsulReactiveDiscoveryClient.java
```diff
@@ -0,0 +1,130 @@
...
+					.hasText(properties.getAclToken())
+							? this.client.getHealthServices(serviceId,
+									this.properties.getDefaultQueryTag(),
+									this.properties.isQueryPassing(), QueryParams.DEFAULT,
+									properties.getAclToken())
+							: this.client.getHealthServices(serviceId,
+									this.properties.getDefaultQueryTag(),
+									this.properties.isQueryPassing(),
+									QueryParams.DEFAULT);
+			return services == null ? Flux.empty()
```
ryanjbaxter: In what situation is `services` `null`?

TYsewyn: Since the `ConsulClient` is working with sub interfaces to handle specific tasks and there is nothing which stops you from returning null if you create a custom implementation, I'd rather have this null check here.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/reactive/ConsulReactiveDiscoveryClientConfiguration.java
```diff
@@ -0,0 +1,80 @@
...
+
+	@Bean
+	@ConditionalOnMissingBean
+	public ConsulReactiveDiscoveryClient consulReactiveDiscoveryClient(
+			ConsulClient client, ConsulDiscoveryProperties discoveryProperties) {
+		return new ConsulReactiveDiscoveryClient(client, discoveryProperties);
+	}
+
+	@Bean
+	@ConditionalOnClass(ReactiveHealthIndicator.class)
```
spencergibb: String attr or nested class

PR #599, opened 2019-10-23
add option to configure Consul check status considered as healthy

jackhammer2k:
Instead of just considering check status "passing" as healthy we make the status configurable. That allows us to also consider e.g. "warning" as healthy.

Default is the current behavior that only allows "passing".

### DIFF
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
@@ -17,11 +17,15 @@
 package org.springframework.cloud.consul.discovery;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import com.ecwid.consul.v1.ConsistencyMode;
+import com.ecwid.consul.v1.health.model.Check;
 
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.context.properties.ConfigurationProperties;
@@ -159,6 +163,12 @@ public class ConsulDiscoveryProperties {
 	 * check passing to the server.
 	 */
 	private boolean queryPassing = false;
+
+	/**
+	 * List of status considered as healthy. Contains "passing" by default.
+	 */
+	private Set<Check.CheckStatus> statusConsideredAsHealthy = new HashSet(
+			Collections.singleton(Check.CheckStatus.PASSING));
 
 	/** Register as a service in consul. */
 	private boolean register = true;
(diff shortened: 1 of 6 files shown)


csabakos:
+1 for this issue. It would be awesome to get this merged :)

jackhammer2k:
Any comment from the maintainers? Does anyone know how to fix the failed build?

TYsewyn:
It looks like your fork is some commits behind our master. Can you please merge our changes into your feature branch and try again?

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulServerList.java
```diff
@@ -83,20 +85,24 @@ public void initWithNiwsConfig(IClientConfig clientConfig) {
...
 	 * @param healthServices the initial list of servers from Consul. Guaranteed to be
 	 * non-empty list
+	 * @param statusConsideredAsHealthy list of status that are considered as
+	 * healthy/passing
 	 * @return ConsulServer instances
-	 * @see ConsulServer#ConsulServer(HealthService)
+	 * @see ConsulServer#ConsulServer(HealthService, java.util.Set)
 	 */
-	protected List<ConsulServer> transformResponse(List<HealthService> healthServices) {
+	protected List<ConsulServer> transformResponse(List<HealthService> healthServices,
```
ryanjbaxter: To avoid changing the method signature I would use `this.properties` in the method rather than pass in the `Set`.  You can add an overloaded method if you want to give the option to past in the `Set`

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulServer.java
```diff
@@ -33,11 +34,15 @@
 
 	private final HealthService service;
 
+	private final Set<Check.CheckStatus> statusConsideredAsHealthy;
+
 	private final Map<String, String> metadata;
 
-	public ConsulServer(final HealthService healthService) {
+	public ConsulServer(final HealthService healthService,
```
ryanjbaxter: We need to maintain backwards compatibility so we need to leave this constructor as is (perhaps mark it as deprecated) and add a new one.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
```diff
@@ -160,6 +164,12 @@
 	 */
 	private boolean queryPassing = false;
 
+	/**
+	 * List of status considered as healthy. Contains "passing" by default.
+	 */
+	private Set<Check.CheckStatus> statusConsideredAsHealthy = new HashSet(
```
ryanjbaxter: We should document the new property

TYsewyn: You can add the explanation of this feature to `docs/src/main/asciidoc/spring-cloud-consul.adoc`.
`docs/src/main/asciidoc/_configprops.adoc` will be automatically generated for you so you don't need to make any changes here.

jackhammer2k:
Looks like there is a sporadically failing test (I did not touch the failing module). Could you please restart the tests?

PR #638, opened 2020-05-03
fix(shellScript): download correct consul zip for arm64 processor

maaaace:
ci_install_consul.sh just downloads the consul for amd64, which cannot be executed on arm64 architecture processors.
Add achitecture checking in this script to support downloading correct binary file.

Signed-off-by: maaaace <fuheming@huawei.com>


spencergibb:
This is for our ci platform which is never arm

spencergibb:
migrated to testcontainers

Issue #646, opened 2020-06-23
ConsulHealthIndicator should not query the entire service catalog


csabakos:
**Describe the bug**
`ConsulHealthIndicator` does a blanket query for the entire service catalog:
```
		final Response<Map<String, List<String>>> services = this.consul
				.getCatalogServices(QueryParams.DEFAULT);
```

In our system, the response JSON is over 500KB long, so this operation is unnecessarily heavy.

Consider removing this check (since maybe the leader check is sufficient), or replacing with a lighter query, such as `/v1/catalog/datacenters`.

spencergibb:
PRs welcome

PR #651, opened 2020-07-08
Allow ConsulHealthIndicator to bypass the services query

bono007:
Allows consumers to decide whether or not to include the services query as part of the healthcheck w/ a default of true - to maintain current functionality.

Fix gh-646.

Enhanced the tests in this area as well:
* Auto-configuration integration test using ApplicationContextRunner
* ConsulHealthIndicator unit test

Enhanced the docs around Consul HTTP check vs. Consul actuator health indicator
* I believe the docs were not completely accurate - confusing at best. 
* I split out the differences between Consul HTTP check and the instances Actuator indicators


docs/src/main/asciidoc/spring-cloud-consul.adoc
```diff
@@ -31,7 +31,7 @@ To activate Consul Service Discovery use the starter with group `org.springframe
 
 === Registering with Consul
 
-When a client registers with Consul, it provides meta-data about itself such as host and port, id, name and tags.  An HTTP https://www.consul.io/docs/agent/checks.html[Check] is created by default that Consul hits the `/health` endpoint every 10 seconds.  If the health check fails, the service instance is marked as critical.
+When a client registers with Consul, it provides meta-data about itself such as host and port, id, name and tags.  An HTTP https://www.consul.io/docs/agent/checks.html[Check] is created by default that Consul hits the `/actuator/health` endpoint every 10 seconds.  If the health check fails, the service instance is marked as critical.
```
bono007, line 34: The default is `/actuator/health`, not `/health`

docs/src/main/asciidoc/spring-cloud-consul.adoc
```diff
@@ -149,21 +149,71 @@ spring.cloud.consul.discovery.management-suffix
...
 spring:
   cloud:
     consul:
       discovery:
-        healthCheckPath: ${management.server.servlet.context-path}/health
+        healthCheckPath: ${management.server.servlet.context-path}/actuator/health
         healthCheckInterval: 15s
 ----
 
-You can disable the health check by setting `management.health.consul.enabled=false`.
```
bono007, line 166: This disables the ConsulHealthIndicator, not the HTTP check from Consul to the instance.

docs/src/main/asciidoc/spring-cloud-consul.adoc
```diff
@@ -149,21 +149,71 @@ spring.cloud.consul.discovery.management-suffix
...
+You can disable the HTTP health check entirely by setting `spring.cloud.consul.discovery.register-health-check=false`.
+
+===== Applying Headers
```
bono007, line 172: Moved the "headers" section under the HTTP check as that is what it belongs to.

docs/src/main/asciidoc/spring-cloud-consul.adoc
```diff
@@ -149,21 +149,71 @@ spring.cloud.consul.discovery.management-suffix
...
+===== ConsulHealthIndicator
+An indicator is configured that verifies the health of the `ConsulClient`.
+
+By default, it retrieves the Consul leader node status and all registered services.
+In deployments that have many registered services it may be costly to retrieve all services on every health check.
+To skip the service retrieval and only check the leader node status set `spring.cloud.consul.health-indicator.include-services-query=false`.
+
+To disable the indicator set `spring.cloud.consul.health-indicator.enabled=false`.
+
+WARNING: When the application runs in https://cloud.spring.io/spring-cloud-commons/2.2.x/reference/html/#the-bootstrap-application-context[bootstrap context mode] (the default),
```
bono007, line 215: This is an interesting caveat - it typically has no effect on the overall actuator health as it registers in the bootstrap context via ConsulAutoRegistration and does not make it to the "health" endpoint. However,  I did not want to "not" document this indicator, as it does register in non-bootstrap mode or if user registers it manually.

spring-cloud-consul-core/src/main/java/org/springframework/cloud/consul/ConsulAutoConfiguration.java
```diff
@@ -82,9 +82,13 @@ public ConsulEndpoint consulEndpoint(ConsulClient consulClient) {
 
 		@Bean
 		@ConditionalOnMissingBean
-		@ConditionalOnEnabledHealthIndicator("consul")
-		public ConsulHealthIndicator consulHealthIndicator(ConsulClient consulClient) {
-			return new ConsulHealthIndicator(consulClient);
+		@ConditionalOnProperty(
+				value = { "spring.cloud.consul.health-indicator.enabled",
+						"management.health.consul.enabled" },
```
bono007, line 87: I followed suit to add the enabled property on the new config props as the DiscoveryClientHealthIndicator is done that way. However, I also respect the typical actuator health enabled property of "`management.health.ID.enabled`". 

Not sure if we want to 
1. keep both
2. only use the Actuator "management.." one
3. use the new one and find a way to deprecate usage of the other (no way other than docs I think though).

spring-cloud-consul-core/src/main/java/org/springframework/cloud/consul/ConsulHealthIndicatorProperties.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.core.style.ToStringCreator;
+
+/**
+ * Configuration properties for {@link ConsulHealthIndicator}.
+ *
+ * @author Chris Bono
+ */
+@ConfigurationProperties("spring.cloud.consul.health-indicator")
```
bono007, line 27: I wanted to go w/ `@ConstructorBinding` but I stuck w/ the consistency of the data/get/set approach.

spring-cloud-consul-core/src/test/java/org/springframework/cloud/consul/ConsulHealthIndicatorTest.java
```diff
@@ -0,0 +1,118 @@
...
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * Unit tests for {@link ConsulHealthIndicator}.
+ *
+ * @author Chris Bono
+ */
+@RunWith(SpringRunner.class)
+public class ConsulHealthIndicatorTest {
```
bono007, line 48: I wanted to test the variants that were added in this code proposal. I first looked at doing in the integration tests w// similar mocks (`@MockBean`) but it became clear they did not belong there. I attempted to balance the level of detail to just mock and test the bare minimum we need to avoid a super brittle test that is in lockstep w/ the indicator impl. I think its not too bad and worth the investment.

spring-cloud-consul-core/src/main/java/org/springframework/cloud/consul/ConsulAutoConfiguration.java
spencergibb: This conditional shouldn't change. It should be `@ConditionalOnEnabledHealthIndicator("consul")`.

spring-cloud-consul-core/src/main/java/org/springframework/cloud/consul/ConsulHealthIndicatorProperties.java
```diff
@@ -0,0 +1,60 @@
...
+public class ConsulHealthIndicatorProperties {
+
+	private boolean enabled = true;
```
spencergibb, line 30: this field isn't needed.

spring-cloud-consul-core/src/test/java/org/springframework/cloud/consul/ConsulAutoConfigurationTests.java
```diff
@@ -1,64 +0,0 @@
...
-
-import static org.assertj.core.api.Assertions.assertThat;
-
-@RunWith(SpringRunner.class)
-@SpringBootTest(properties = { "spring.cloud.consul.tls.key-store-instance-type=JKS",
-		"spring.cloud.consul.tls.key-store-path=src/test/resources/server.jks",
-		"spring.cloud.consul.tls.key-store-password=letmein",
-		"spring.cloud.consul.tls.certificate-path=src/test/resources/trustStore.jks",
-		"spring.cloud.consul.tls.certificate-password=change_me" })
-public class ConsulAutoConfigurationTests {
```
spencergibb, line 42: looks like you renamed to `ConsulAutoConfigurationTest`, please revert.

spring-cloud-consul-core/src/test/java/org/springframework/cloud/consul/ConsulHealthIndicatorDownTest.java
```diff
@@ -30,6 +30,8 @@
 import static org.assertj.core.api.Assertions.assertThat;
 
 /**
+ * Integration test for {@link ConsulHealthIndicator}.
```
spencergibb, line 33: Unrelated change

spring-cloud-consul-core/src/test/java/org/springframework/cloud/consul/ConsulHealthIndicatorUpTest.java
```diff
@@ -30,6 +30,8 @@
 import static org.assertj.core.api.Assertions.assertThat;
 
 /**
+ * Integration test for {@link ConsulHealthIndicator}.
```
spencergibb, line 33: Unrelated change

spring-cloud-consul-core/src/main/java/org/springframework/cloud/consul/ConsulHealthIndicatorProperties.java
```diff
@@ -0,0 +1,60 @@
...
+
+	private boolean includeServicesQuery = true;
```
spencergibb, line 32: If this was targeted to the 2.2.x branch, we could do 'true' there and then in master, since it is a major release, change to 'false'.

spring-cloud-consul-core/src/test/java/org/springframework/cloud/consul/ConsulHealthIndicatorTest.java
bono007: I think I can simplify this by using the existing integration up/down test and adding `@SpyBean/@MockBean`. I will see how it looks in code and then may update when I make the other suggested changes.

spring-cloud-consul-core/src/main/java/org/springframework/cloud/consul/ConsulHealthIndicatorProperties.java
bono007: Good idea. I will adjust.

bono007:
@spencergibb - thx for the feedback. I made the suggested updates. 

I  also combined the important cases in the unit test w/ the already existing up/down tests. Feels like a good middle ground that exercises the primary cases and minimizes the level of mocking.

I will target 2.2 once we are ok w/ the current modifications. Thanks.

spring-cloud-consul-core/src/main/java/org/springframework/cloud/consul/ConsulHealthIndicatorProperties.java
bono007: If this is a new property in master only, could we not just make the default `false` there and not put in `2.2.x`?

bono007: I went ahead and targeted 2.2.x in https://github.com/spring-cloud/spring-cloud-consul/pull/653 . I believe this can be closed in favor of that.

spencergibb:
Closing in favor of #653

PR #653, opened 2020-07-12
Allow ConsulHealthIndicator to bypass the services query

bono007:
**NOTE:** This is the 2.2.x target of https://github.com/spring-cloud/spring-cloud-consul/pull/651 (which can probably be closed at this point).

Allows consumers to decide whether or not to include the services query as part of the healthcheck w/ a default of true - to maintain current functionality.

Fixes gh-646.

Enhanced the tests in this area as well:
- Auto-configuration integration test using ApplicationContextRunner
- ConsulHealthIndicator unit test

Enhanced the docs 
- Split out the differences between Consul HTTP check and the instances Actuator indicators

### DIFF
--- a/docs/src/main/asciidoc/_configprops.adoc
+++ b/docs/src/main/asciidoc/_configprops.adoc
@@ -77,3 +77,5 @@
 |spring.cloud.consul.tls.key-store-instance-type |  | Type of key framework to use.
 |spring.cloud.consul.tls.key-store-password |  | Password to an external keystore.
 |spring.cloud.consul.tls.key-store-path |  | Path to an external keystore.
+
+|===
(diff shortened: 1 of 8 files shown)


docs/src/main/asciidoc/_configprops.adoc
```diff
@@ -77,4 +78,4 @@
 |spring.cloud.consul.tls.key-store-password |  | Password to an external keystore.
 |spring.cloud.consul.tls.key-store-path |  | Path to an external keystore.
 
-|===
\ No newline at end of file
+|===
```
bono007, line 81: Sorry for all the chatter and back and forth on this - having issue w/ formatting and not sure if my formatting is correct and it was previously checked in w/o trailing spaces removed.

bono007: I have since learned that this is auto-generated in `spring-cloud-commons` - I am assuming the same here. I will remove my updates and move the docs to the javadoc of the properties

spencergibb:
Squashed and merged via 6d6d6fccac046eca96ed97c12ec20749b31ea7e7

PR #657, opened 2020-07-24
Adds a probe implementation for ConsulDiscoveryClient

bono007:
This is the motivation for https://github.com/spring-cloud/spring-cloud-commons/issues/785 and relies on https://github.com/spring-cloud/spring-cloud-commons/pull/789 .


spring-cloud-consul-binder/src/main/java/org/springframework/cloud/consul/binder/ConsulSendingHandler.java
```diff
@@ -50,9 +50,8 @@ protected void handleMessageInternal(Message<?> message) {
 		}
 
 		Object payload = message.getPayload();
-		if (payload instanceof byte[])
-		{
-			payload = Arrays.toString((byte[])payload);
+		if (payload instanceof byte[]) {
```
bono007, line 53: *UNRELATED* Formatter changed this and I left the adjustment in as it was out of line w/ code style previously

Issue #682, opened 2020-11-10
Removal of "tags as metadata" breaks filtering service lookup by dynamic tag value


bono007:
In [this commit](https://github.com/bono007/spring-cloud-consul/commit/c3f23675499ca74b6a716bc61b733d14824fea39#diff-2a667f48e0dcb665886b85796597ff947f181bba055c1216d573b68cb3764a8c) the "tags as metadata" ability was removed. 

This was used in 2 areas (that I know of):
1) During service registration
2) During service lookup

This is a breaking change for us due to #2.

Our services are not registered via Spring Cloud Consul Discovery but we do use it to lookup services in consul (#2 above). We have 1000s of service instances that do not use metadata and instead use a tag to identify the "name" of the service instance. We find all service instances and then we spin through and filter those w/ the desired tag value (in the consul metadata map). This quit working as the metadata map is now empty. 

### Here are some options that come to mind:

1) Update all of our services to properly use metadata when they are registered
   - This is a non-starter and out of our control based on our deployment/infra landscape.

2) Set the `spring.cloud.consul.discovery.default-query-tag: myServiceTag` 
     - This works for a static tag value, but we use it to lookup a service dynamically. If we could pass the query tag in [here](https://github.com/spring-cloud/spring-cloud-consul/blob/master/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClient.java#L67), that would work - with the one caveat that it would be specific to the consul impl and not in the DiscoveryClient contract (which is ok for our case).

3) Re-instate the ability to use "tags as metadata" for use case #2

4) Re-instate the "tags as metadata" feature but default it to "off"

I am willing to submit a merge request for whatever option the team lands on.

bono007:
@spencergibb can you take a scan of this ticket? This will be a blocker for us to upgrade to `2020.0.0`.

spencergibb:
Indeed, 2020.0 was a chance for us to make breaking changes. Tags as metadata was a hack/workaround.

>  We find all service instances and then we spin through and filter those w/ the desired tag value

Since that is your current mode of operations, I've created a `ConsulServiceInstance` class that you may cast to that contains a `getTags()` method and a `getHealthService()` if there is any other data anyone might need that isn't in the interface.

bono007:
Thanks @spencergibb - this will work. I do also like having the ability to get at the backing health service if any other "need" arises later. 

[...]

Issue #680, opened 2020-11-03
`spring.cloud.consul.discovery.server-list-query-tags` is ignored when using Spring Cloud LoadBalancer


mario45211:
**Describe the bug**
Configuration under `spring.cloud.consul.discovery.server-list-query-tags` property is ignored when quering service's instances while using Spring Cloud LoadBalancer.

**Sample**
When application has given configuration (which enables Spring LoadBalancer):
```
spring:
  cloud:
    consul:
      discovery:
        server-list-query-tags:
          <my-service-id>: leader
    loadbalancer:
      ribbon:
        enabled: false
```
there is expecting to load-balance requests between <my-service-id> instances only with `leader` tag, but this tag is ignored instead and requests are dispatched between all available instances.

In case of usage Ribbon (`spring.loadbalancer.ribbon.enabled=true`) above issue does not occur.

bono007:
I noticed that property was not used internally by `spring-cloud-consul` during my workaround for #682 and actually suggested a rework/rename of this property.

OlgaMaciaszek:
For now, let's get a fix of this issue with the old property - if we end up changing it later on, it'll be easy to refactor and also not an issue since we have not yet done the release and 2020.0.0 is going to be a major.

PR #684, opened 2020-11-23
Adds support for multiple default query tags

bono007:
### Summary
Adds support for multiple default query tags and deprecates other unused related properties.

Workaround for gh-682

### Details
This code proposal is based on [option #2](https://github.com/spring-cloud/spring-cloud-consul/issues/682#issue-739940435) from the related issue:
> 2. Set the spring.cloud.consul.discovery.default-query-tag: myServiceTag

It is a meet-in-the-middle solution that solves my application's need while still leaving the metadata removal in-place. Additionally it improves the current state of the query tag configuration. 

| ℹ️ |
| --- | 
After adding the tests for this I realized that the multiple tags are combined (AND'ed) by the ConsulHealthClient (as one would expect they would be). This does not help solve my application's needs as we would have to achieve and `OR` of the filters. "Give me all services with tag 'east' or 'west'" for example. That being said, I feel like this is still an improvement in the query tags configuration and I have consolidated the query tags tests into a single `ApplicationContextRunner` based integration test with more thorough coverage.

| ℹ️ I will update the docs shortly |
| --- |

@spencergibb or team, please advise one way or the other when you can. I know you are all super busy trying to get `2020.0` released - thank you very much for all of the hard work going into that.


spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
```diff
@@ -159,19 +164,36 @@
 
 	/**
 	 * Map of serviceId's -> tag to query for in server list. This allows filtering
-	 * services by a single tag.
+	 * services by a single tag. (Deprecated - instead use 'serviceInstanceQueryTags')
 	 */
+	@Deprecated
 	private Map<String, String> serverListQueryTags = new HashMap<>();
```
bono007, line 170: So this property is not used internally by anything. If someone configures this - it does nothing. I deprecated it along w/ its associated default property `defaultQueryTag`. I am not sure the proper way to deprecate config props in Spring Cloud (in Boot there is the `@DeprecatedConfigurationProperty` but its not in use here).

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
```diff
@@ -159,19 +164,36 @@
...
+	 */
+	private List<String> defaultServiceInstanceQueryTags;
+
 	/**
 	 * Map of serviceId's -> datacenter to query for in server list. This allows looking
 	 * up services in another datacenters.
 	 */
 	private Map<String, String> datacenters = new HashMap<>();
 
-	/** Tag to query for in service list if one is not listed in serverListQueryTags. */
```
bono007, line 172: Just moved this to be close to its counterpart.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
```diff
@@ -159,19 +164,36 @@
...
+	 * (Deprecated - instead use 'defaultServiceInstanceQueryTags')
+	 */
+	@Deprecated
+	private String defaultQueryTag;
+
+	/**
+	 * Map of serviceId -> tags to use when querying for service instances. This allows
+	 * filtering service instances by tags.
+	 */
+	private Map<String, List<String>> serviceInstanceQueryTags = new HashMap<>();
```
bono007, line 183: This pair is the new replacement - named accordingly for what it is being used for (as well as I had to deal w/ the close names of the deprecated pair).

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
```diff
@@ -219,13 +241,41 @@ public ConsulDiscoveryProperties(InetUtils inetUtils) {
 
 	/**
 	 * @param serviceId The service who's filtering tag is being looked up
-	 * @return The tag the given service id should be filtered by, or null.
+	 * @return The tags the given service id should be filtered by, or null.
+	 * @deprecated instead use {@link #getQueryTagsForService(String)}
 	 */
+	@Deprecated
 	public String getQueryTagForService(String serviceId) {
```
bono007, line 248: While this method attempts to first use the service specific query tag, its not used by anything internally other than tests. I deprecated it in favor of the following plural version.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
```diff
@@ -219,13 +241,41 @@ public ConsulDiscoveryProperties(InetUtils inetUtils) {
...
+	 * @return non-null list of tags to filter the service instances by (empty if no tags
+	 * configured)
+	 */
+	public List<String> getQueryTagsForService(String serviceId) {
+		// Use specific config entry if available
+		if (this.serviceInstanceQueryTags.containsKey(serviceId)) {
+			return Collections.unmodifiableList(this.serviceInstanceQueryTags.get(serviceId));
+		}
+
+		// Honor use of deprecated property while giving tip to consuming application
```
bono007, line 268: We honor the deprecated `defaultQueryTag` so that if consuming apps were relying on that to feed into the services query, it still will do that. 

I chose to not honor the elder `serverListQueryTags` since there is no internal use of it. IOW - this will not break anyone that is setting that property. The only way that property has any impact is if consumers are also using the deprecated singular `getQueryTagForService` for some other external use. When/if consumers that are setting `serverListQueryTags` decide to move off of it they can also move off of the deprecated singular accessor `getQueryTagForService`.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
OlgaMaciaszek: That's the way we deprecate it, however, please make sure you also add `deperecated` tag and link to the new version in the javadocs whenever you deprecate items.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
```diff
@@ -159,19 +164,36 @@
...
+	/**
+	 * Map of serviceId -> tags to use when querying for service instances. This allows
+	 * filtering service instances by tags.
+	 */
+	private Map<String, List<String>> serviceInstanceQueryTags = new HashMap<>();
+
+	/**
+	 * Tags to use when querying for service instances if the server id in question does
+	 * not have an entry in 'serviceInstanceQueryTags'
+	 */
```
OlgaMaciaszek, line 188: Use `@link` instedad of '' in javadocs.

OlgaMaciaszek (commented):
Hi, @bono007, thanks for the PR. Other than some changes in the Javadoc, the code looks good to me. However, I will leave it to @spencergibb for the final review, since he has more knowledge on the context of [these changes](https://github.com/bono007/spring-cloud-consul/commit/c3f23675499ca74b6a716bc61b733d14824fea39#).

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
bono007: I usually do that but have learned to **not** do that when in Spring Boot as the post processor that generates the docs from the `@ConfigurationProperties` does not handle that well (IIRC). Is that not the case here?

spencergibb:
At this point we can no longer have breaking changes. How about reusing the existing property, but allowing a comma separated list to be parsed and passed to consul?

bono007:
> At this point we can no longer have breaking changes. How about reusing the existing property, but allowing a comma separated list to be parsed and passed to consul?

Oh man! No more breaking stuff? :) I missed the window. 

So, not sure if you saw the updated description comment but I figured out that the tags are combined anyways - which is not the behavior my application needed. This is not a problem though as you have already provided a workaround for that w/ the drop-down to the `ConsulServiceInstance` and I am planning on using that. 

I mean, we could delete this pull request right now as my problem is already solved. That being said, I totally don't mind adjusting this to parse the tag property (csv) which will add this functionality in but w/o breaking changes on the properties. Another option is deleting all of the added functionality in here and keeping the integration test consolidation. 

Just let me know. I have zero problem with deleting code I just wrote. It's not the first I wrote, it won't be the last.  Hehe.

spencergibb:
If you'd like to updating to parsing a comma separated list, great, otherwise we can close this.

bono007:
Cool.  I will update it later today.

spencergibb:
No rush. We're all going on holiday until after the new year :-)

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClient.java
```diff
@@ -67,13 +68,10 @@ public String description() {
...
 	private void addInstancesToList(List<ServiceInstance> instances, String serviceId, QueryParams queryParams) {
 		HealthServicesRequest.Builder requestBuilder = HealthServicesRequest.newBuilder()
-				.setPassing(this.properties.isQueryPassing()).setQueryParams(queryParams)
-				.setToken(this.properties.getAclToken());
-		String queryTag = this.properties.getQueryTagForService(serviceId);
-		if (queryTag != null) {
-			requestBuilder.setTag(queryTag);
-		}
+				.setPassing(properties.isQueryPassing()).setQueryParams(queryParams).setToken(properties.getAclToken());
+		Optional.ofNullable(properties.getQueryTagsForService(serviceId)).ifPresent(requestBuilder::setTags);
```
bono007, line 72: The new `getQueryTagsForService` behaves identical to the previously used `getQueryTagForService` in the single tag value case. *IFF* there are multiple tag property values does it diverge (by adding multiple tags).

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
```diff
@@ -218,14 +224,44 @@ public ConsulDiscoveryProperties(InetUtils inetUtils) {
...
+	 * Whitespace will be trimmed off each entry.
+	 * @param serviceId the service whose instances are being looked up
+	 * @return the array of tags to filter the service instances by - it will be null if
+	 * no tags are configured for the service and the default query tag is not configured
+	 * or if a single tag is configured and it is the empty string
+	 */
+	@Nullable
+	public String[] getQueryTagsForService(String serviceId) {
+		String queryTagStr = getQueryTagForService(serviceId);
+		if (queryTagStr == null || queryTagStr.isEmpty()) {
```
bono007, line 259: This allows us to knockout a previously set default during tests or in a profile specific property override.  Imagine the case where the default profile yaml section sets 
```java
spring.cloud.consul.discovery.default-query-tag: tag1
```
but we want to disable the tag filtering in qa env (or in a test) with 
```java
spring.cloud.consul.discovery.default-query-tag:
```
Prior to this this change, that would not work.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/reactive/ConsulReactiveDiscoveryClient.java
```diff
@@ -75,11 +76,14 @@ public String description() {
 	}
 
 	private List<HealthService> getHealthServices(String serviceId) {
-		HealthServicesRequest request = HealthServicesRequest.newBuilder().setTag(this.properties.getDefaultQueryTag())
```
bono007, line 78: I just realized that the Reactive client was still suffering from #680 . This brings it inline w/ the new behaviour.

spring-cloud-consul-discovery/src/test/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClientQueryTagsTests.java
```diff
@@ -0,0 +1,215 @@
...
+		service.setAddress("localhost");
+		service.setId(String.format("%s-%s-%s", NAME, env, region));
+		service.setName(NAME);
+		service.setPort(port);
+		service.setTags(Arrays.asList(env, region));
+		return service;
+	}
+
+	@Test
+	void singleMatchingTagSpecifiedOnDefaultQueryTagProperty() {
```
bono007, line 71: I am using Jupiter rather than Junit4 here. The class/method visibility modifiers reflect that as do the method names (not having to be prefixed w/ `test`)

spring-cloud-consul-discovery/src/test/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClientQueryTagsTests.java
```diff
@@ -0,0 +1,215 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+
+import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
+
+import com.ecwid.consul.v1.ConsulClient;
+import com.ecwid.consul.v1.agent.model.NewService;
+import org.junit.jupiter.api.Test;
```
bono007, line 27: I am using Jupiter rather than Junit4 here. The class/method visibility modifiers reflect that as do the method names (not having to be prefixed w/ `test`)

spring-cloud-consul-discovery/src/test/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClientQueryTagsTests.java
```diff
@@ -0,0 +1,215 @@
...
+
+	@Test
+	void emptyTagSpecifiedOnDefaultQueryTagProperty() {
+		appContextRunner.withPropertyValues("spring.cloud.consul.discovery.default-query-tag=")
+				.run(context -> assertThatGetInstancesReturnsExpectedServices(context, QA_WEST_SERVICE, QA_EAST_SERVICE,
+						PROD_WEST_SERVICE, PROD_EAST_SERVICE));
+	}
+
+	@Test
+	void singleTagSpecifiedOnServerListQueryTagsProperty() {
```
bono007, line 108: I attempted to reduce overlap w/ the `ConsulDiscoverPropertiesTest` as much as possible. However there is still overlap. 

One approach to reduce the overlap would be to mock out the `CDP.getQueryTagsForService` to return  whatever is needed rather than have the same group of tests for the default query tag and the server list query tag values. Ideally, this test should **not** concern itself w/ the mechanics of `CDP#getQueryTagsForService`. That being said, there is still value in having these here as they do integration test the `@ConfigurationProperties` aspect of CDP. 

I realized I was bikeshedding and when I do that while writing tests, I typically err on the side of extra coverage rather than thinning it out.

spring-cloud-consul-discovery/src/test/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryPropertiesTests.java
```diff
@@ -19,15 +19,20 @@
 import java.util.Collections;
 import java.util.Map;
 
-import org.junit.Before;
-import org.junit.Test;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
```
bono007, line 23: Again, here I converted over to Jupiter. I know we strive to not pull multiple levers in a single pull request. However, the changes were significant enough in this test that I went ahead and bumped up. I feel that if we don't do this a little along the way that JUnit4 will stay in here for quite some time.

spring-cloud-consul-discovery/src/test/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClientQueryTagsTests.java
```diff
@@ -0,0 +1,215 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Lightweight integration tests to verify Consul query tags usage in
+ * {@link ConsulDiscoveryClient#getInstances(String)}.
+ *
+ * @author Chris Bono
```
bono007, line 45: This test supersedes and incorporates the same coverage as the 2 deleted tests in this commit. What do we do in the case where we incorporate someone's code (`@author Piotr Wielgolaski` in this case) into a new replacement? Should I move the `@author` tag to the new file? I would not care either way if it were my author tag but other's may have a different opinion on that.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClient.java
spencergibb, line 72: We don't use `Optional` in place of if statements, please update.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java
```diff
@@ -218,14 +224,44 @@ public ConsulDiscoveryProperties(InetUtils inetUtils) {
...
+			return null;
+		}
+		return Arrays.stream(queryTagStr.split(",")).map(String::trim).toArray(String[]::new);
```
spencergibb, line 262: Change to `StringUtils.commaDelimitedListToStringArray()`.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/reactive/ConsulReactiveDiscoveryClient.java
```diff
@@ -75,11 +76,14 @@ public String description() {
...
-				.setPassing(this.properties.isQueryPassing()).setQueryParams(QueryParams.DEFAULT)
-				.setToken(this.properties.getAclToken()).build();
+		HealthServicesRequest.Builder requestBuilder = HealthServicesRequest.newBuilder()
+				.setPassing(properties.isQueryPassing()).setQueryParams(QueryParams.DEFAULT)
+				.setToken(properties.getAclToken());
+		Optional.ofNullable(properties.getQueryTagsForService(serviceId)).ifPresent(requestBuilder::setTags);
```
spencergibb, line 82: Again, replace with if statement.

spring-cloud-consul-discovery/src/test/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClientQueryTagsTests.java
spencergibb: Yes, please

spencergibb (changes requested):
LGTM with a few small changes.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClient.java
bono007: K. Will do. I can see in this simple case that this could be an "abuse" of an Optional (aka replacing a simple if w/ it). 

What is the stance when accessing an N level, nullable property path such as `user.contactInfo.address.street`:

**If statement**
```java
User.contactInfo.address.street
if (user != null && user.getContactInfo() != null && user.getContactInfo().getAddress() != null) {
  return user.getContactInfo().getAddress().getStreet();
}
```
vs. using `Optional + map` such as:
```java
return Optional.ofNullable(user)
   .map(User::contactInfo)
   .map(ContactInfo::getAddress)
   .map(Address::street)
    .orElse(null);
```
In groovy and other langs this would be as simple as `user?.contactInfo?.address?.street` of course. 

Not asking to spark a debate or push any preference - just want to know if this is a hard rule or not.

spencergibb: In Spring Cloud we don't use optional as an if statement at all

bono007:
@spencergibb I will get the updates in w/in the next 24hrs.

bono007:
@spencergibb - I think I addressed all requested changes. Thanks.

spring-cloud-consul-discovery/src/test/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryPropertiesTests.java
```diff
@@ -114,18 +114,6 @@ void getTagsReturnsMultipleFromServerListMapEntry() {
 		assertThat(properties.getQueryTagsForService(SERVICE_NAME_IN_MAP)).containsExactly("foo", "bar");
 	}
 
-	@Test
```
bono007, line 63: Note that the StringUtils method does not remove whitespace. This is fine - being too nice w/ the property values is not worth the added code and users can just enter the values in w/o whitespace around entries.

spencergibb: `tokenizeToStringArray` does that for you

spencergibb:
You mentioned some documentation updates?

spring-cloud-consul-discovery/src/test/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryPropertiesTests.java
bono007: Ok I will check that out.

bono007:
> You mentioned some documentation updates?

I updated [ConsulDiscoveryProperties](https://github.com/spring-cloud/spring-cloud-consul/pull/684/files#diff-c8106df333831cc0cadbb5401a1de8d1b9788aeb8180c22d179706b31c62962dR240) but will make sure any mention of the query tags int he docs gets an update as well.

bono007:
> > You mentioned some documentation updates?
> 
> I updated [ConsulDiscoveryProperties](https://github.com/spring-cloud/spring-cloud-consul/pull/684/files#diff-c8106df333831cc0cadbb5401a1de8d1b9788aeb8180c22d179706b31c62962dR240) but will make sure any mention of the query tags int he docs gets an update as well.

@spencergibb I think `_configprops.adoc` is. the best place for the new property doc as its the only mention of the default query tag capability I could find in the docs.  However, I noticed that file is not getting updated automatically like I thought it would. Do I need to do something to kick that off? Is it broken? Or does it happen during a later build step?

bono007:
@spencergibb this one still shows as `Changes requested` but I think I addressed all concerns. Please take a scan when you can. Thanks.

spencergibb (approved):

PR #685, opened 2020-11-24
Respect the 'serverListQueryTags' on service instance lookup

bono007:
Respect the 'spring.cloud.consul.discovery.server-list-query-tags' property during service instance lookup

Fixes gh-680

### DIFF
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClient.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClient.java
@@ -40,6 +40,7 @@ import static org.springframework.cloud.consul.discovery.ConsulServerUtils.getMe
  * @author Spencer Gibb
  * @author Joe Athman
  * @author Tim Ysewyn
+ * @author Chris Bono
  */
 public class ConsulDiscoveryClient implements DiscoveryClient {
 
@@ -77,11 +78,14 @@ public class ConsulDiscoveryClient implements DiscoveryClient {
 
 	private void addInstancesToList(List<ServiceInstance> instances, String serviceId,
 			QueryParams queryParams) {
-
-		HealthServicesRequest request = HealthServicesRequest.newBuilder()
-				.setTag(this.properties.getDefaultQueryTag())
+		HealthServicesRequest.Builder requestBuilder = HealthServicesRequest.newBuilder()
 				.setPassing(this.properties.isQueryPassing()).setQueryParams(queryParams)
-				.setToken(this.properties.getAclToken()).build();
+				.setToken(this.properties.getAclToken());
+		String queryTag = this.properties.getQueryTagForService(serviceId);
+		if (queryTag != null) {
+			requestBuilder.setTag(queryTag);
+		}
+		HealthServicesRequest request = requestBuilder.build();
 		Response<List<HealthService>> services = this.client.getHealthServices(serviceId,
 				request);
 
--- a/spring-cloud-consul-discovery/src/test/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClientServerListQueryTagTests.java
+++ b/spring-cloud-consul-discovery/src/test/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClientServerListQueryTagTests.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright 2013-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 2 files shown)


spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClient.java
```diff
@@ -73,10 +73,14 @@ public String description() {
...
 
 	private void addInstancesToList(List<ServiceInstance> instances, String serviceId, QueryParams queryParams) {
-
-		HealthServicesRequest request = HealthServicesRequest.newBuilder().setTag(this.properties.getDefaultQueryTag())
+		HealthServicesRequest.Builder requestBuilder = HealthServicesRequest.newBuilder()
 				.setPassing(this.properties.isQueryPassing()).setQueryParams(queryParams)
-				.setToken(this.properties.getAclToken()).build();
+				.setToken(this.properties.getAclToken());
+		String queryTag = this.properties.getQueryTagForService(serviceId);
+		if (queryTag != null) {
```
bono007, line 80: Before we were setting this unconditionally (even when null) so what was getting passed into Consul health client was `tags=[null]` but they must handle that properly. I went ahead and only set when its specified.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClient.java
```diff
@@ -73,10 +73,14 @@ public String description() {
 	}
 
 	private void addInstancesToList(List<ServiceInstance> instances, String serviceId, QueryParams queryParams) {
-
-		HealthServicesRequest request = HealthServicesRequest.newBuilder().setTag(this.properties.getDefaultQueryTag())
+		HealthServicesRequest.Builder requestBuilder = HealthServicesRequest.newBuilder()
 				.setPassing(this.properties.isQueryPassing()).setQueryParams(queryParams)
-				.setToken(this.properties.getAclToken()).build();
+				.setToken(this.properties.getAclToken());
+		String queryTag = this.properties.getQueryTagForService(serviceId);
```
bono007, line 79: To live up to the "contract" in the docs, 1st the `serverListQueryTags` should be consulted, then the `defaultQueryTag`. This method already existed and was not currently being used.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClient.java
```diff
@@ -73,10 +73,14 @@ public String description() {
 	}
 
 	private void addInstancesToList(List<ServiceInstance> instances, String serviceId, QueryParams queryParams) {
-
-		HealthServicesRequest request = HealthServicesRequest.newBuilder().setTag(this.properties.getDefaultQueryTag())
+		HealthServicesRequest.Builder requestBuilder = HealthServicesRequest.newBuilder()
```
OlgaMaciaszek, line 76: Please add your with the `@author` tag to the javadoc.

spring-cloud-consul-discovery/src/test/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClientServerListQueryTagTests.java
```diff
@@ -0,0 +1,112 @@
...
+	@Autowired
+	private ConsulClient consulClient;
+
+	private NewService intgService = serviceForEnvironment("intg", 9081);
+
+	private NewService uatService = serviceForEnvironment("uat", 9080);
+
+	@Before
+	public void setUp() {
+		this.consulClient.agentServiceRegister(this.intgService);
```
OlgaMaciaszek, line 75: Please do not use `this` where unnecessary. I know it's confusing, cause we have it elsewhere, but we're trying to modify it and at least not add it in new code.

OlgaMaciaszek (changes requested):
Hi @bono007, thanks a lot for this PR. It looks good - have just requested some cosmetic changes.
Since the issue is also present in `Hoxton`, could you submit this PR against the `2.2.x` branch instead of `master` though?

spring-cloud-consul-discovery/src/test/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClientServerListQueryTagTests.java
bono007: I am not a fan of `this` personally so I am excited about that. Is there a plan to update the formatter? Also, is this a "Spring-wide" initiative, or "Spring Cloud"? I just want to know because as I bounce around projects (Spring Boot in particular) it will be confusing to keep track of who does/not expect it.

bono007:
@OlgaMaciaszek I targeted 2.2.x and made the suggested updates.

spring-cloud-consul-discovery/src/test/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClientServerListQueryTagTests.java
OlgaMaciaszek: Unfortunately, it's just a Spring Cloud initiative. Our formatting differs slightly from team to team. In fact, you should not get formatting issues for that while building cloud projects. If you are, please let me know - we'll need to fix this.

OlgaMaciaszek (approved):
LGTM.

Issue #197, opened 2016-06-14
Re-register service in consul if the service get's de-registered


mirwais27:
If a local consul agent is restarted (for what ever reason) then the local Spring Boot Apps (with spring-cloud-consul inside) does not a re-connect. Anything special that I missed to configure for having this?

spencergibb:
I just tried it and works ok. There are no persistent connections to consul, so the next time it tries to connect, it should work. Are you seeing any errors?

mirwais27:
We are starting the consul agent like that: `consul.exe agent -log-level info -data-dir .\consul-data -server -ui -ui-dir=.\consul-ui -bootstrap -advertise=127.0.0.1`. We configured also the retry feature in pur spring boot apps as mentioned here: http://cloud.spring.io/spring-cloud-static/spring-cloud.html#spring-cloud-consul-retry 
The issue is that from time to time we update or reinstall the consul agent - with deleting the **consul-data** folder. And if we start then the consul agent then the spring boot apps can't reconnect: The consul-ui `http://localhost:8500/` shows only the consul agent itself, but not the running services anymore. Workaround is to restart all spring boot apps as well. But if you have many of them then it's really annoying. 

It would be good if the spring boot apps re-register again automatically if consul is running again and they are not registered yet. Would make spring boot apps even more resilient.

spencergibb:
> with deleting the consul-data folder
is the key statement there. So you're not asking for a reconnection, that works. You're asking for reregistration.

mirwais27:
As you describe: re-register even if consul agent is started with a fresh empty consul-data folder would be really nice!

spencergibb:
Knowing when to re-register is the problem then.

pwielgolaski:
I have the same problem, when my consul instance is restarted, spring boot app is sending heartbeat bot service is missing and it get 500.
It would be nice to have re-register feature

spencergibb:
how do you know a 500 should reregister?

pwielgolaski:
Fair point, we can also check if error message contains information about missing check 
https://github.com/hashicorp/consul/blob/master/command/agent/acl.go#L372

avarabyeu:
@spencergibb i believe, if health check attempt fails, we need to check whether service is still registered in consul. So we can re-register service if it got de-registered somehow.

FlowerWrong:
The same issue.

spencergibb:
[...]

PR #690, opened 2020-12-26
TTL check uses actuator for application status

bono007:
#### Summary
Adds support for the TTL check to take into account the current application health status.

#### How does it get the app status?
If the consuming application is a Spring Boot Actuator application with its health endpoint enabled/available then the TTL check will use that to determine system health. Otherwise, the previous behavior of `report always passing to Consul agent` will be used. There is also the ability to configure your own custom status provider.

|  |
| :--- |
| :information_source:   It would be a good idea to read the updates to the docs at this point as it calls out some other configuration and extension options. |

| Healthcheck | Consul UI |
| --- | --- |
|  |  |
| | |
| | |
|  |  |

PR #691, opened 2021-01-03
Add option for re-registration when ephemeral Consul is restarted

making:
Fixes gh-197 .

I haven't updated the doc as TTL is not yet documented.

### DIFF
--- a/docs/src/main/asciidoc/_configprops.adoc
+++ b/docs/src/main/asciidoc/_configprops.adoc
@@ -33,7 +33,8 @@
 |spring.cloud.consul.discovery.health-check-url |  | Custom health check url to override default.
 |spring.cloud.consul.discovery.heartbeat.enabled | `false` | 
 |spring.cloud.consul.discovery.heartbeat.interval-ratio |  | 
-|spring.cloud.consul.discovery.heartbeat.ttl | `30s` | 
+|spring.cloud.consul.discovery.heartbeat.ttl | `30s` |
+|spring.cloud.consul.discovery.heartbeat.re-register-service-on-failure | `false` | Enables service re-registration when the heartbeat fails.
 |spring.cloud.consul.discovery.hostname |  | Hostname to use when accessing server.
 |spring.cloud.consul.discovery.include-hostname-in-instance-id | `false` | Whether hostname is included into the default instance id when registering service.
 |spring.cloud.consul.discovery.instance-group |  | Service instance group.
--- a/docs/src/main/asciidoc/_configprops.adoc
+++ b/docs/src/main/asciidoc/_configprops.adoc
@@ -34,7 +34,7 @@
 |spring.cloud.consul.discovery.heartbeat.enabled | `false` | 
 |spring.cloud.consul.discovery.heartbeat.interval-ratio |  | 
 |spring.cloud.consul.discovery.heartbeat.ttl | `30s` |
-|spring.cloud.consul.discovery.heartbeat.re-register-service-on-failure | `false` | Enables service re-registration when the heartbeat fails.
+|spring.cloud.consul.discovery.heartbeat.reregister-service-on-failure | `false` | Enables service re-registration when the heartbeat fails.
 |spring.cloud.consul.discovery.hostname |  | Hostname to use when accessing server.
 |spring.cloud.consul.discovery.include-hostname-in-instance-id | `false` | Whether hostname is included into the default instance id when registering service.
 |spring.cloud.consul.discovery.instance-group |  | Service instance group.
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/HeartbeatProperties.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/HeartbeatProperties.java
@@ -51,7 +51,7 @@ public class HeartbeatProperties {
 	@DecimalMax("0.9")
 	private double intervalRatio = 2.0 / 3.0;
 
-	private boolean reRegisterServiceOnFailure = false;
+	private boolean reregisterServiceOnFailure = false;
 
 	/**
 	 * @return the computed heartbeat interval
@@ -92,12 +92,12 @@ public class HeartbeatProperties {
 		this.intervalRatio = intervalRatio;
 	}
 
-	public boolean isReRegisterServiceOnFailure() {
-		return this.reRegisterServiceOnFailure;
+	public boolean isReregisterServiceOnFailure() {
+		return this.reregisterServiceOnFailure;
 	}
 
-	public void setReRegisterServiceOnFailure(boolean reRegisterServiceOnFailure) {
-		this.reRegisterServiceOnFailure = reRegisterServiceOnFailure;
+	public void setReregisterServiceOnFailure(boolean reregisterServiceOnFailure) {
+		this.reregisterServiceOnFailure = reregisterServiceOnFailure;
 	}
 
 	@Override
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/HeartbeatProperties.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/HeartbeatProperties.java
@@ -51,6 +51,8 @@ public class HeartbeatProperties {
 	@DecimalMax("0.9")
 	private double intervalRatio = 2.0 / 3.0;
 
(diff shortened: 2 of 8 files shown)


making:
I found the doc about TTL is included in #690

pcornelissen:
I have not checked your PR, but I just wanted to document my idea somewhere:
A good indicator if the service needs to register in consul (again) is that when it fetches the instances list from consul and itself is not included in that list.
The neat thing is that this criteria would remove the necessity to register programatically on startup, because the service can just connect to consul and start fetching the instance list and then (obviously) it's not present yet and the criteria above would trigger a registration.

Just my 2C, I'm not sure if that is feasible or the best way

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/HeartbeatProperties.java
```diff
@@ -51,6 +51,8 @@
 	@DecimalMax("0.9")
 	private double intervalRatio = 2.0 / 3.0;
 
+	private boolean reRegisterServiceOnFailure = false;
```
spencergibb, line 54: `reregisterServiceOnFailure`

docs/src/main/asciidoc/_configprops.adoc
```diff
@@ -33,7 +33,8 @@
 |spring.cloud.consul.discovery.health-check-url |  | Custom health check url to override default.
 |spring.cloud.consul.discovery.heartbeat.enabled | `false` | 
 |spring.cloud.consul.discovery.heartbeat.interval-ratio |  | 
-|spring.cloud.consul.discovery.heartbeat.ttl | `30s` | 
+|spring.cloud.consul.discovery.heartbeat.ttl | `30s` |
+|spring.cloud.consul.discovery.heartbeat.re-register-service-on-failure | `false` | Enables service re-registration when the heartbeat fails.
```
spencergibb, line 37: `reregister`

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ReRegistrationPredicate.java
```diff
@@ -0,0 +1,40 @@
...
+package org.springframework.cloud.consul.discovery;
+
+import com.ecwid.consul.v1.OperationException;
+
+/**
+ * Predicate on whether to re-register service.
+ *
+ * @author Toshiaki Maki
+ */
+public interface ReRegistrationPredicate {
```
spencergibb, line 26: Why new interface? Just use `Predicate<OperationException`.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/TtlScheduler.java
```diff
@@ -77,22 +87,46 @@ public void remove(String instanceId) {
...
+			}
+			catch (OperationException e) {
+				if (this.ttlScheduler.heartbeatProperties.isReRegisterServiceOnFailure()
+						&& this.ttlScheduler.reRegistrationPredicate.isEligible(e)) {
+					log.warn(e.getMessage());
+					if (this.ttlScheduler.registeredService != null) {
+						if (log.isInfoEnabled()) {
+							log.info("Re-register " + this.ttlScheduler.registeredService);
+						}
+						this.ttlScheduler.client.agentServiceRegister(this.ttlScheduler.registeredService,
```
spencergibb, line 120: `this.client`?

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/serviceregistry/ConsulServiceRegistry.java
```diff
@@ -68,7 +68,7 @@ public void register(ConsulRegistration reg) {
 			NewService service = reg.getService();
 			if (this.heartbeatProperties.isEnabled() && this.ttlScheduler != null && service.getCheck() != null
 					&& service.getCheck().getTtl() != null) {
-				this.ttlScheduler.add(reg.getInstanceId());
+				this.ttlScheduler.add(reg.getService());
```
spencergibb, line 71: Why this change?

spencergibb (changes requested):
Looks like a good start.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ReRegistrationPredicate.java
making: I tend to avoid using classes under `java.*` package as a bean type. It can potentially conflict with other libraries. For example, Spring Cloud Function uses `Function`, `Consumer` and `Supplier`. If there is no possibility of a conflict in `Predicate`, I don't mind dropping this class.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/TtlScheduler.java
making: `client` is not visible in `ConsulHeartbeatTask`

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/serviceregistry/ConsulServiceRegistry.java
making: `ConsulClient.agentServiceRegister` requires `NewService` instance rather than just a service id.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/HeartbeatProperties.java
making: done

docs/src/main/asciidoc/_configprops.adoc
making: done

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/TtlScheduler.java
```diff
@@ -42,27 +43,36 @@
...
+		this.heartbeatProperties = heartbeatProperties;
+		this.discoveryProperties = discoveryProperties;
 		this.client = client;
+		this.reregistrationPredicate = reregistrationPredicate;
 	}
 
-	@Deprecated
 	public void add(final NewService service) {
 		add(service.getId());
+		this.registeredService = service;
```
spencergibb, line 66: There's a reason add adds into a map. Any reason why `NewService` doesn't get added to one as well.

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/TtlScheduler.java
spencergibb: but isn't that because you changed the class to `static`?

making: yeah I've change it to static because of the testability.
Would you like me to revert to private and use `ReflectionUtils` in `ConsulHeartbeatTaskTests`?

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/TtlScheduler.java
making: Do you meath it should have `Map<String, NewService> registeredServices` ?

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/TtlScheduler.java
spencergibb: No

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/TtlScheduler.java
spencergibb: Yes, I think so. You'll need to update remove as well.

making: Updated

PR #744, opened 2021-08-13
Remove extra / character at the beginning of the context

### LINKED ISSUE #730: Declaring a custom context in spring.config.import leads to error caused by an extra leading '/'
Hi,

Every microservice we run can access its configuration on Consul under a folder conventioned as `$product/$environment/$module/$app/$version`

Example:
`my-product/my-env/my-module/my-app/2.0.0`

To configure our Spring Boot apps, we set on them an env var:
`SPRING_CONFIG_IMPORT=consul:/my-product/my-env/my-module/my-app/0.0.0`

Unfortunately our configuration is never overridden with values from Consul because of a bug in [ConsulPropertySource](https://github.com/spring-cloud/spring-cloud-consul/blob/77416811bf9f2698c717fb325cf9355c3ec04ab7/spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulPropertySource.java#L101)

```
// context = /my-product/my-env/my-module/my-app/0.0.0
// key       =  my-module/my-env/my-module/my-app/0.0.0/sample/nested/my-super-prop1
// please pay attention to the leading '/' which is the cause of all the troubles....

// So the Line 101 key = key.replace(this.context, "") doesn't strip the context as expected!

protected void parsePropertiesInKeyValueFormat(List<GetValue> values) {
		if (values == null) {
			return;
		}

		for (GetValue getValue : values) {
			String key = getValue.getKey();
			if (!StringUtils.endsWithIgnoreCase(key, "/")) {
				key = key.replace(this.context, "").replace('/', '.');
				String value = getValue.getDecodedValue();
				this.properties.put(key, value);
			}
		}
	}
```

Previously (Spring Boot `2.4.5`) we used a workaround by defining a `;` in the String:
`SPRING_CONFIG_IMPORT=consul:localhost:8500;my-product/my-env/my-module/my-app/0.0.0`

But is does not work anymore on greater versions of Spring Boot/Spring Cloud .
We now get a `java.lang.NumberFormatException: For input string: "8500;"`

Is that something that can be directly fixed in Spring Cloud Consul or is there an other solution to get it working?

fmartinou:
Fixes #730

### DIFF
--- a/spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulPropertySource.java
+++ b/spring-cloud-consul-config/src/main/java/org/springframework/cloud/consul/config/ConsulPropertySource.java
@@ -63,6 +63,9 @@ public class ConsulPropertySource extends EnumerablePropertySource<ConsulClient>
 		if (!this.context.endsWith("/")) {
 			this.context = this.context + "/";
 		}
+		if (this.context.startsWith("/")) {
+			this.context = this.context.substring(1);
+		}
 
 		Response<List<GetValue>> response = this.source.getKVValues(this.context, this.configProperties.getAclToken(),
 				QueryParams.DEFAULT);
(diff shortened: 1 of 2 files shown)


ryanjbaxter (changes requested):
Can you provide a unit test for this?

fmartinou:
> Can you provide a unit test for this?

- I added a trivial UT to validate that the context is rewritten as expected (removing first slash...)
- I updated existing UTs so test data contains a context with the extra slash

ryanjbaxter (approved):

Issue #751, opened 2021-09-09
ConsulReactiveDiscoveryClient supports fault when consul cluster unavailable


chenzhiguo:
`ConsulReactiveDiscoveryClient` should support the backup service list function. If the consul cluster is not available, the data should be returned to the last successful request.

ryanjbaxter:
Looks like you have a potential fix in your fork, would you like to submit that as a PR to the project?

chenzhiguo:
Yeah, I have submit the PR(#752).

spencergibb:
See comment here https://github.com/spring-cloud/spring-cloud-consul/pull/752#issuecomment-1462561443

PR #752, opened 2021-09-09
Support service discovery list backup. (#751)

chenzhiguo:
Cache the service list data, if the Consul cluster is not available, use cache data to ensure that the service is highly available.

### DIFF
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClient.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClient.java
@@ -19,6 +19,7 @@ package org.springframework.cloud.consul.discovery;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import com.ecwid.consul.v1.ConsulClient;
 import com.ecwid.consul.v1.QueryParams;
@@ -27,6 +28,8 @@ import com.ecwid.consul.v1.catalog.CatalogServicesRequest;
 import com.ecwid.consul.v1.health.HealthServicesRequest;
 import com.ecwid.consul.v1.health.model.HealthService;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.cloud.client.ServiceInstance;
 import org.springframework.cloud.client.discovery.DiscoveryClient;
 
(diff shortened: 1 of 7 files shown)


spencergibb:
Can you describe the use case here? Both spring cloud loadbalancer and ribbon have caches for loadbalancing. If we decide to do this we should be consistent with s-c-lb https://github.com/spring-cloud/spring-cloud-commons/tree/main/spring-cloud-loadbalancer/src/main/java/org/springframework/cloud/loadbalancer/cache

chenzhiguo:
@spencergibb I found that the cache in the load balancer will be lost after a certain period of time. This is not the same as the purpose of my PR. Of course, we can modify the the load balancer cache to achieve the same purpose. 
I don't quite clear what other help do I do?

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClient.java
```diff
@@ -73,11 +81,20 @@ private void addInstancesToList(List<ServiceInstance> instances, String serviceI
...
-		for (HealthService service : services.getValue()) {
-			instances.add(new ConsulServiceInstance(service, serviceId));
+		try {
+			Response<List<HealthService>> services = this.client.getHealthServices(serviceId, request);
+
+			for (HealthService service : services.getValue()) {
+				instances.add(new ConsulServiceInstance(service, serviceId));
+			}
+			if (properties.isBackup()) {
+				serviceInstanceBackup.put(serviceId, instances);
```
ryanjbaxter, line 91: Shouldn't you purge the "backup" before adding the new instances?  You could end up with stale data in the backup.

ryanjbaxter (changes requested):

spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryClient.java
chenzhiguo: I think that the same serviceId, after being added, the old instances will be covered by new data, isn't it? @ryanjbaxter

ryanjbaxter: What if the service was removed?

chenzhiguo: I think that when the service is removed, consul can also accept the request. The data inside the Backup will be replaced by a empty list. What should this be reasonable? But if consul does not provide a service normally, it should return the last acquisition of the service instance list, no matter whether it is empty.

ryanjbaxter: Sorry i thought this code was iterating all services and not instances of a given service.

chenzhiguo: Yes, but if your assumption, `instances` will be an empty list when there is no service instance returns, which will also be placed in the map. isn't it?

ryanjbaxter: Yes correct, I don't think there is a problem after all

chenzhiguo:
Oh...Is there anything unsuitable?

spencergibb:
If the loadbalancer cache looses information, that should be fixed. I've marked #751 for discussion.

spencergibb:
You can change the TTL of the cache to something very large here https://docs.spring.io/spring-cloud-commons/docs/current/reference/html/#loadbalancer-cache-configuration

Otherwise, if you think there is a bug in spring cloud loadbalancer, open an issue in https://github.com/spring-cloud/spring-cloud-commons/issues

PR #773, opened 2022-03-21
Fixed the problem of creating ConsulServiceInstance

manzhizhen:
When I registered a "myApp" service on consumer, because I didn't set the port number, spring cloud consumer discovery threw a NullPointerException when reading the service and creating a DefaultServiceInstance. Later, I found that it was because the port of the DefaultServiceInstance constructor was of type "int" rather than "Integer", so when our port was null, the reading of the service would fail, Therefore, I think spring cloud consumer discovery should filter out such illegal instances and ensure that other instances of the service are still read successfully.

Here I attach the exception stack information:

```
java.lang.NullPointerException: null
	at org.springframework.cloud.consul.discovery.ConsulServiceInstance.<init>(ConsulServiceInstance.java:37) ~[spring-cloud-consul-discovery-3.0.4.jar:3.0.4]
	at org.springframework.cloud.consul.discovery.reactive.ConsulReactiveDiscoveryClient.lambda$getInstances$0(ConsulReactiveDiscoveryClient.java:69) ~[spring-cloud-consul-discovery-3.0.4.jar:3.0.4]
	at reactor.core.publisher.FluxDefer.subscribe(FluxDefer.java:46) ~[reactor-core-3.4.14.jar:3.4.14]
```

Here is also a screenshot of the Consul service registration:
image

image

### DIFF
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/reactive/ConsulReactiveDiscoveryClient.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/reactive/ConsulReactiveDiscoveryClient.java
@@ -16,26 +16,25 @@
 
 package org.springframework.cloud.consul.discovery.reactive;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-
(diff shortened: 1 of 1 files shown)


spencergibb:
bb8e674acbc677409d19bba85c48db23df2b0a8e made it so port is never null so NPE is avoided.

PR #786, opened 2022-06-01
Do not supply discovery client when Consul discovery client disabled

### LINKED ISSUE #785: Properties eureka.client.enabled and spring.cloud.consul.discovery.enabled have no effects without bootstrap
**Describe the bug**
I try to locate the config server by using a discovery service client (Eureka or Consul). When only one artifact `spring-cloud-starter-netflix-eureka-client` or `spring-cloud-starter-consul-discovery` is present in the classpath, the app locates successfully the config server, regardless the value of `eureka.client.enabled` and `spring.cloud.consul.discovery.enabled` (check out the `main` or `feature/register_with_consul` branches of the sample project).

When both artifacts `spring-cloud-starter-netflix-eureka-client` or `spring-cloud-starter-consul-discovery` are present in the classpath, the Consul discovery client is always used by the app to locate the config server, regardless the value of `eureka.client.enabled` and `spring.cloud.consul.discovery.enabled` (check out the `feature/without_bootstrap` branch of the sample project).

On the other hand, it works as expected when I use the Config First Bootstrap with Discovery Lookup First enabled (`feature/with_bootstrap` branch of the sample project).

**Sample**
Here is the link to the sample project [https://github.com/xhuy-le/spring-cloud-discovery-client-issues](https://github.com/xhuy-le/spring-cloud-discovery-client-issues).

ryanjbaxter:
Fixes #785

### DIFF
--- a/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConditionalOnConsulDiscoveryEnabled.java
+++ b/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConditionalOnConsulDiscoveryEnabled.java
@@ -36,7 +36,12 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
 @Inherited
-@ConditionalOnProperty(value = "spring.cloud.consul.discovery.enabled", matchIfMissing = true)
+@ConditionalOnProperty(value = ConditionalOnConsulDiscoveryEnabled.PROPERTY, matchIfMissing = true)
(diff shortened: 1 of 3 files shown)


ryanjbaxter:
@spencergibb let me know what you think about this change....the other option is to continue to return `null` but then we will also need fix `ConfigServerInstanceProvider` to handle the `null` function.  Ultimately the end result is the same...an `IllegalStateException` will be thrown withe the same error message indicating not being able to find the config server.

spencergibb (approved):

xhuy-le:
@ryanjbaxter I'm wondering if we should extend these changes to other discovery client implementations that could suffer the same problem (Zookeeper, Kubernetes, ...) ?

PR #803, opened 2022-12-10
Use testcontainers' consul module

eddumelendez:
Replace `GenericContainer` by `ConsulContainer` and provide
logback file to display proper testcontainers logs.


spencergibb:
I just did some rework around Testcontainers which caused a conflict here, so I went ahead and did it myself. Thanks for letting me know about the module!

I also kept logging as is.

PR #808, opened 2023-02-12
Added status code 404 to ReregistrationPredicate

attermann:
Consul agent as of approximately v1.11.2 returns status code 404 on agent health check instead of status code 500. This change resulted in re-registration not taking place when expected. Adding status code 404 to the ReregistrationPredicate enables re-registration with both older and newer versions of Consul agent.


spencergibb:
Thanks for the contribution. Added a test in 6378b72526f1205d7634d58b0d7e9645622f1410

Issue #475, opened 2018-12-11
Consider replacing https://github.com/Ecwid/consul-api with Interface Clients


spencergibb:
A move to non-blocking infrastructure like spring-cloud-loadbalancer will be problematic with a blocking client.

varnson:
Is anyone  working on consul-api with WebClient?

pturczyk:
It seems that although [consula-api](https://github.com/Ecwid/consul-api) is still listed on [official consul API docs](https://developer.hashicorp.com/consul/api-docs/libraries-and-sdks), it is not really actively maintained. Last release happened on Apr 9, 2020 and reported [Issues](https://github.com/Ecwid/consul-api/issues) are not being addressed nor commented. 

Perhaps it's worth bringing up the topic of replacing the underlying Consul API again. Especially, as some API's has been[ deprecated](https://github.com/Ecwid/consul-api/issues/246) and some new features like [caching](https://developer.hashicorp.com/consul/api-docs/features/caching) are not supported.

spencergibb:
I would use the new [Interface Clients](https://docs.spring.io/spring-framework/reference/integration/rest-clients.html#rest-http-interface) from Spring Framework. It would make it easy to do both reactive and blocking.

alex-kormukhin:
https://github.com/spring-cloud/spring-cloud-consul/issues/809

M-Whitaker:
Is anyone working on this? Are you accepting PRs?

spencergibb:
Not yet, PRs wecome

M-Whitaker:
I’m happy to start to look at creating an interface client as suggested removing the lib from the config module.

M-Whitaker:
@spencergibb Any comments on the current draft PR are most welcome to get some feedback on the implementation.

spencergibb:
I will look when I have time

monwolf:
@spencergibb it's still pending, are you planning to review this PR?

spencergibb:
@monwolf as the PR stands, it will need to wait for our major in 2025.1 (Nov 2025) as it is a breaking change. If you want to make it opt-in, I can schedule it for 2025.0 (May 2025) and then we can remove the old one in 2025.1.

PR #840, opened 2024-08-07
Replace Consul HTTP Client with Interface Clients

M-Whitaker:
As stated in #475, Consul token query parameter deprecated in Consul 1.17 however the current Consul HTTP client seems to be no longer maintained.

Moving to interface clients also has the added benefit for allowing for blocking and non-blocking clients to be used.

Further to this change, the docker image used for integration testing this lib is using an outdated image: https://hub.docker.com/_/consul and does not have the latest consul images so this needed to be moved to https://hub.docker.com/r/hashicorp/consul.

Any suggestions are VERY welcome, I assume this only as a first draft for this change so have only migrated the config module at this moment.

### DIFF
--- a/spring-cloud-consul-core/src/main/java/org/springframework/cloud/consul/IConsulClient.java
+++ b/spring-cloud-consul-core/src/main/java/org/springframework/cloud/consul/IConsulClient.java
@@ -19,6 +19,9 @@ package org.springframework.cloud.consul;
 import java.util.List;
 import java.util.Map;
 
+import org.springframework.cloud.consul.model.http.agent.Service;
+import org.springframework.cloud.consul.model.http.catalog.CatalogService;
+import org.springframework.cloud.consul.model.http.catalog.Node;
 import org.springframework.web.service.annotation.GetExchange;
 
 public interface IConsulClient {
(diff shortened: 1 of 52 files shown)


spring-cloud-consul-core/src/main/java/org/springframework/cloud/consul/IConsulClient.java
```diff
@@ -0,0 +1,83 @@
...
+			@RequestHeader(value = ACL_TOKEN_HEADER, required = false) String aclToken);
+
+	@GetExchange("/v1/kv/{context}?recurse")
+	ResponseEntity<List<GetValue>> getKVValues(@PathVariable String context,
+			@RequestHeader(value = ACL_TOKEN_HEADER, required = false) String aclToken);
+
+	@GetExchange("/v1/kv/{context}?recurse")
+	ResponseEntity<List<GetValue>> getKVValues(@PathVariable String context,
+			@RequestHeader(value = ACL_TOKEN_HEADER, required = false) String aclToken,
+			@RequestParam("wait") @WaitTimeFormat String waitTime, @RequestParam("index") long index);
```
M-Whitaker, line 72: @WaitTimeFormat was an attempt to use a `Long` for waitTime and have spring convert it into a string like `3s`. I'd appreciate help understanding how to autoconfigure this behaviour

spencergibb:
Thanks @M-Whitaker for your patience and what a huge contribution!

M-Whitaker:
Hi @spencergibb I had seen this as a first draft however no longer have the capacity to support this change. If you want to use it as a starter for ten great but I think this would still need quite a lot of work

spencergibb:
All the tests pass, so, that works for me. We'll take it from here, thanks again

PR #846, opened 2025-02-07
NOISSUE: Fix documentation link for Consul agent

varunu28:
Link for Consul agent on https://spring.io/projects/spring-cloud-consul routes to a 404
image

This code change fixes the link & updates it to use https://developer.hashicorp.com/consul/docs/agent#starting-the-consul-agent for Consul agent


spencergibb:
We should delete that file since it is not the source of truth for the website. This has been fixed https://spring.io/projects/spring-cloud-consul

